PR #2, opened 2026-02-12
Add first iteration of WCI client

02strich:
## What was changed
This adds the WCI client that will be consumed by the Frontend and Matching services to implement the WCI-related APIs.

## Why?
Matching and Frontend service shouldn't know about the implementation details of WCI so this abstracts it

Re https://temporalio.atlassian.net/browse/COM-6

---

> [!NOTE]
> **Medium Risk**
> Touches workflow update/query behavior and introduces new signal emission from matching events, which can impact workflow throughput and error handling if misconfigured or noisy.
> 
> **Overview**
> Adds a new `wci/client` package that encapsulates interacting with Worker Controller Instance workflows via query/list/update/delete operations, including atomic *update-with-start* behavior, retry/throttling logic, and namespace-level instance limit enforcement.
> 
> Extends the WCI workflow interface to include typed compute provider/scaling configuration fields and a `SignalTaskMatch` payload, and wires the client as a `TaskMatchListener` that batches task-match signals and emits new error metrics when signaling fails. Also updates the WCI workflow’s query/memo and update validation to include the new scaling configuration and `BuildId` field naming.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0e518b6e3f3d264bd4df5b3c345cefc55af23321. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/wci/client/client.go
+++ b/wci/client/client.go
@@ -1,0 +1,488 @@
+// Package client contains the client to interact with Worker Controller Instance workflows
+package client
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	commonpb "go.temporal.io/api/common/v1"
+	deploymentpb "go.temporal.io/api/deployment/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/serviceerror"
+	updatepb "go.temporal.io/api/update/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/managed-workers/wci/workflow/iface"
+	"go.temporal.io/server/api/historyservice/v1"
+	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/backoff"
+	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/persistence/visibility/manager"
+	"go.temporal.io/server/common/sdk"
+	"go.temporal.io/server/common/testing/testhooks"
+	"go.temporal.io/server/common/worker_versioning"
+)
+
+const (
+	WorkerControllerPerNSWorkerTaskQueue = "temporal-sys-worker-controller-per-ns-tq"
+
+	// minSignalIntervalNoSyncMatch is the minimum time between SignalWorkflowExecution calls per (namespace, workflow ID) for no-sync matches.
+	minSignalIntervalNoSyncMatch = 500 * time.Millisecond
+	// minSignalIntervalSyncMatch is the minimum time between SignalWorkflowExecution calls per (namespace, workflow ID) for sync matches.
+	minSignalIntervalSyncMatch = 1 * time.Minute
+)
+
+type (
+	Client interface {
+		WorkerControllerInstanceExists(
+			ctx context.Context,
+			namespaceEntry *namespace.Namespace,
+			version *deploymentpb.WorkerDeploymentVersion,
+		) (bool, error)
+
+		DescribeWorkerControllerInstance(
+			ctx context.Context,
+			namespaceEntry *namespace.Namespace,
+			version *deploymentpb.WorkerDeploymentVersion,
+		) (*iface.QueryDescribeWorkerControllerInstanceResponse, []byte, error)
+
+		ListWorkerControllerInstances(
+			ctx context.Context,
+			namespaceEntry *namespace.Namespace,
+			pageSize int,
+			nextPageToken []byte,
+		) ([]*WorkerControllerInstanceSummary, []byte, error)
+
+		UpdateWorkerControllerInstance(
+			ctx context.Context,
+			namespaceEntry *namespace.Namespace,
+			version *deploymentpb.WorkerDeploymentVersion,
+			conflictToken []byte,
+			identity string,
+			computeProviderDetails *iface.ComputeProviderDetails,
+			scalingConfiguration *iface.ScalingConfiguration,
+		) error
+
+		DeleteWorkerControllerInstance(
+			ctx context.Context,
+			namespaceEntry *namespace.Namespace,
+			version *deploymentpb.WorkerDeploymentVersion,
+			identity string,
+		) error
+
+		SignalTaskAddEvent(
+			ctx context.Context,
+			namespaceEntry *namespace.Namespace,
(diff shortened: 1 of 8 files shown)


02strich:
bugbot run

02strich:
@cursor review

02strich:
@cursor review

wci/client/util.go
```diff
@@ -0,0 +1,209 @@
...
+	})
+	if err != nil {
+		var notFound *serviceerror.NotFound
+		if errors.As(err, &notFound) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	// Deployment exists only if the entity wf is running
```
jaypipes: "Deployment"? is this just copy/pasta?

wci/client/client.go
```diff
@@ -0,0 +1,611 @@
...
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/persistence/visibility/manager"
+	"go.temporal.io/server/common/sdk"
+	"go.temporal.io/server/common/tasks"
+	"go.temporal.io/server/common/testing/testhooks"
+	"go.temporal.io/server/common/worker_versioning"
+)
+
+const (
+	WorkerControllerPerNSWorkerTaskQueue = "temporal-sys-worker-controller-per-ns-tq"
```
jaypipes, line 35: it's not clear to me what this task queue will be used for.

wci/workflow/iface/wf_iface.go
```diff
@@ -47,60 +51,114 @@ var WorkerControllerInstanceVisibilityBaseListQuery = fmt.Sprintf(
...
-		BuildID        string
+	ScalingConfiguration struct {
+		ScalingAlgorithm ScalingAlgorithmType `json:"scaling_algorithm,omitempty"`
+	}
 
-		State *WorkerControllerInstanceLocalState
+	QueueTypeScalingMetrics struct {
+		LastWorkerStart int64 `json:"last_worker_start,omitempty"`
+
+		LastQueueDeptch    int64   `json:"last_queue_depth"`
```
jaypipes:
```suggestion
		LastQueueDepth    int64   `json:"last_queue_depth"`
```

wci/client/workflow_interaction.go
```diff
@@ -0,0 +1,209 @@
...
+
+	var outcome *updatepb.Outcome
+
+	err := backoff.ThrottleRetryContext(ctx, func(ctx context.Context) error {
+		// historyClient retries internally on retryable rpc errors, we just have to retry on
+		// successful but un-completed responses.
+		res, err := historyClient.ExecuteMultiOperation(ctx, multiOpReq)
+		if err != nil {
+			return err
+		}
```
jaypipes, line 195: ugh.

I guess CHASM has a cleaner way of doing this sort of thing than legacy workflows?

wci/client/workflow_interaction.go
```diff
@@ -0,0 +1,209 @@
+package client
```
jaypipes, line 1: Not sure if this was adopted from `temporalio/temporal` codebase, but putting functions into files called `util.go` is an anti-pattern and should be avoided...

02strich:
@cursor review

02strich:
@cursor review

wci/client/workflow_interaction.go
```diff
@@ -0,0 +1,209 @@
...
+	var res *historyservice.QueryWorkflowResponse
+	var err error
+	err = backoff.ThrottleRetryContext(ctx, func(ctx context.Context) error {
+		res, err = historyClient.QueryWorkflow(ctx, req)
+		return err
+	}, retryPolicy, isRetryableQueryError)
+	return res, err
+}
+
+func updateWorkflow(
```
jaypipes, line 72: the UpdateWorkflow operation is convoluted. It doesn't update the workflow or the workflow execution at all. I don't know why it's called UpdateWorkflow instead of something like SignalWithWait or something like that :(

wci/workflow/iface/wf_iface.go
```diff
@@ -47,60 +51,114 @@ var WorkerControllerInstanceVisibilityBaseListQuery = fmt.Sprintf(
...
-		Identity string
+		Identity string `json:"identity,omitempty"`
 	}
 	DeleteWorkerControllerInstanceResponse struct{}
 
+	SignalTaskMatchRequest struct {
+		TaskQueueName string                `json:"task_queue_name"`
+		TaskQueueType enumspb.TaskQueueType `json:"task_queue_type"`
+
+		BacklogDepth   int64   `json:"backlog_count"`
```
jaypipes:
```suggestion
		BacklogCount   int64   `json:"backlog_count"`
```

(to align with the metric of the same name)

jaypipes (commented):
The entity workflows stuff makes the code extremely difficult to read and understand. In my experience, simpler is better.

I don't really understand why this WIC isn't just a very simple program that polls the public-facing DescribeTaskQueue API and its resulting [`stats.approximate_backlog_count` field](https://github.com/temporalio/api/blob/b66af7ab522c34ef697aed6c87b7f69b422c9259/temporal/api/taskqueue/v1/message.proto#L109) which, is noted in the code comments, is reliable to use for scaling events.

wci/client/util.go
02strich: yeah, also a useless comment so removing it

wci/client/client.go
02strich: It's the task queue the WCI worker listens on

wci/client/workflow_interaction.go
02strich: yeah, some of these APIs are not pretty at all. CHASM would be quite a bit nicer (which was why that was the original plan)

wci/client/workflow_interaction.go
02strich: indeed adopted with surprise - restructured now

wci/client/workflow_interaction.go
02strich: It does make an Update call - which in Temporal is a specific thing https://temporal.io/blog/announcing-a-new-operation-workflow-update

Why that is called "Update" I have no idea...

02strich:
>I don't really understand why this WIC isn't just a very simple program that polls the public-facing DescribeTaskQueue API and its resulting [stats.approximate_backlog_count field](https://github.com/temporalio/api/blob/b66af7ab522c34ef697aed6c87b7f69b422c9259/temporal/api/taskqueue/v1/message.proto#L109) which, is noted in the code comments, is reliable to use for scaling events.

Mostly two reasons:
1/ polling for count is too slow for quite a few serverless scenarios (i.e. while backlog is a good metric, it is not sufficient)
2/ the effort required to make such a "simple" program reliable and scaleable

smuneebahmad (approved):

PR #5, opened 2026-02-26
Introduce new spec structure for local state

02strich:
## What was changed
This refactors the LocalState of the workflow to a spec+state pair. The Spec allows configuring different provider and scaler settings for different task type sets. Task Queue name is not a configuration dimension to limit the complexity of the implementation for now.

## Why?
Lays the data structure foundation for the core workflow.


wci/workflow/iface/spec.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// ComputeProviderSpec is a single provider type and its settings (used per task queue type or as default).
+	ComputeProviderSpec struct {
+		ProviderType ComputeProviderType   `json:"provider_type,omitempty"`
+		Settings     ComputeProviderConfig `json:"settings,omitempty"`
+	}
+
+	ScalingAlgorithmType   string
+	ScalingAlgorithmConfig map[string]any
+	ScalingAlgorithmState  map[string]any
```
jaypipes: this one probably shouldn't go in a file called `spec.go` :)

Also, I'm sure you will want to make this into an actual struct instead of a map of string to any, yes? If you do make it into an actual struct, better to name it `Status` instead of `State` if we're trying to match the pattern from Kubernetes API machinery.

wci/workflow/iface/spec.go
```diff
@@ -0,0 +1,251 @@
...
+		return defaultVal
+	}
+	switch val := v.(type) {
+	case int:
+		return int64(val)
+	case int64:
+		return val
+	case float64:
+		return int64(val)
+	default:
```
jaypipes: It's highly likely that the map values will be `string` if, for example, the maps are being generated from something like `json.Unmarshal` or `yaml.Unmarshal`. I know in the validate functions below, you return an error if the value isn't an int or float, but just trying to think ahead for the consumer of this API...

wci/workflow/iface/workflow.go
```diff
@@ -80,23 +67,22 @@ type (
...
 		CreateTime           *timestamppb.Timestamp `json:"create_time,omitempty"`
 		LastModifierIdentity string                 `json:"last_modifier_identity,omitempty"`
 	}
 
 	QueryDescribeWorkerControllerInstanceResponse struct {
-		DeploymentName string                 `json:"deployment_name,omitempty"`
-		BuildId        string                 `json:"build_id,omitempty"`
-		CreateTime     *timestamppb.Timestamp `json:"create_time,omitempty"`
+		DeploymentName    string                 `json:"deployment_name,omitempty"`
+		DeploymentBuildID string                 `json:"deployment_build_id,omitempty"`
```
jaypipes, line 82: 👍

wci/workflow/iface/spec.go
02strich: yeah, moved out and renamed to `Status` as the pattern is too close to ignore

> Also, I'm sure you will want to make this into an actual struct instead of a map of string to any, yes?

Not really for the moment as the Status that gets maintained is quite algo specific and undefined from the outside. If there is overall status that needs to be maintained (more akin to the Status of a k8s deployment) that would be a different type (and probably a struct)

wci/workflow/iface/spec.go
02strich: When do you think it would be a string? (I feel like I am missing context)

Today all the values are numbers so they come back as floats from the `json` package.

jaypipes: if someone does this, for example:

```yaml
scaler:
  minInstances: "5"
```

jaypipes (approved):

wci/workflow/iface/spec.go
02strich: added support for it

jaypipes (approved):
👍

PR #6, opened 2026-02-26
Introduce Compute Providers & Scaling Algorithms

02strich:
## What was changed
This brings in the Compute Provider and Scaling Algorithm abstractions, as well as the workflow and activity code that connects them.

## Why?
Both Compute Providers and algorithms are extension points of the platform and things we'll keep extending so having them abstract is critical.

### DIFF
--- a/wci/workflow/compute_provider/registry.go
+++ b/wci/workflow/compute_provider/registry.go
@@ -1,0 +1,75 @@
+// Package computeprovider contains the implementations of the various compute providers
+package computeprovider
+
+import (
+	"context"
+	"slices"
+	"sync"
+
+	"github.com/temporalio/temporal-managed-workers/wci/client"
+	"github.com/temporalio/temporal-managed-workers/wci/workflow/iface"
+	"go.temporal.io/server/common/dynamicconfig"
+)
+
+type (
+	LaunchStrategy string
+
+	ComputeProviderConstructor func(context.Context, *dynamicconfig.Collection) (ComputeProvider, error)
+
+	ComputeProvider interface {
+		// LaunchStrategy returns the LaunchStrategy used by this ComputeProvider, i.e. whether it is
+		// starts instances that end by themselves (invoke) or manages a managed
+		// set of instances that auto-restart if they fail (worker-set)
+		LaunchStrategy() LaunchStrategy
+
+		// ValidateConfig checks the provided configuration for correctness. This might involved
+		// invoking cloud provider functions to check permissions as well. If any issues are found
+		// returns an error with a description
+		ValidateConfig(ctx context.Context, config iface.ComputeProviderConfig) error
+
+		// InvokeWorker starts a new worker instance when using the LaunchStrategy 'invoke'. any
+		// error is returned if the invocation failed, but not if the invoked work dies before connecting
+		// to Temporal. Returns an error for other launch strategies.
+		InvokeWorker(ctx context.Context, config iface.ComputeProviderConfig) error
+
+		// UpdateWorkerSetSize updates the size of the managed worker set to the provided 'size' when using
+		// the LaunchStrategy 'worker-set'. An error is returned when the size update fails, but might not
+		// if the triggered instance starts/stops fail. Always returns an error for other launch stratgies.
+		UpdateWorkerSetSize(ctx context.Context, config iface.ComputeProviderConfig, size int) error
+	}
+)
+
+const (
+	LaunchStrategyInvoke    LaunchStrategy = "invoke"
+	LaunchStrategyWorkerSet LaunchStrategy = "worker-set"
+)
+
+var (
+	providerConstructorsMu sync.RWMutex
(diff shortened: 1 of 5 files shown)


wci/workflow/scaling_algorithm/registry.go
```diff
@@ -0,0 +1,123 @@
...
+	"go.temporal.io/server/common/dynamicconfig"
+)
+
+type (
+	ActionType string
+
+	ScalingAlgorithmConstructor func(context.Context) (ScalingAlgorithm, error)
+
+	ScalingAction struct {
+		SpecKey string     `json:"spec_key"`
```
jaypipes, line 22: What's this?

wci/workflow/compute_provider/registry.go
```diff
@@ -0,0 +1,52 @@
...
+	"go.temporal.io/server/common/dynamicconfig"
+)
+
+type (
+	ComputeProviderConstructor func(context.Context, *dynamicconfig.Collection) (ComputeProvider, error)
+
+	ComputeProvider interface {
+		ValidateConfig(ctx context.Context, config iface.ComputeProviderConfig) error
+		StartNewWorkerInstance(ctx context.Context, config iface.ComputeProviderConfig, count int) error
+		StopWorkerInstance(ctx context.Context, config iface.ComputeProviderConfig, count int) error
```
jaypipes: Couple points on this interface.

1. If the function of this method is to start potentially multiple instances of a worker, it should be called `StartWorkerInstances`... but, that said...
2. I'm not actually a fan of "batch create" APIs like this. It reminds me a bit of [EC2's RunInstances API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html), which I think is a poor interface because it improperly connotes the resource being created as multiple EC2 instances instead of what is *actually* the resource being created, which is the *set* of identical EC2 instances. In other words, the resource is the group, not the individual members of that group. In fact, EC2's own developers internally disliked this API because the EC2 AutoScalingGroup APIs that succeeded it were better encapsulated/designed and having the multi-instance batch create `RunInstances` API call caused a bunch of unnecessary confusion for developers.

Because of (2), I would recommend making this API single-targeted and call it `StartWorkerInstance`.

The other thing that is unfortunate about the Worker model in Temporal (and not something you can do anything about at this point...) is that there is no stable identifier for a Worker Instance that you can supply to the SDK's `worker.New()` function so that you can, for example, generate a worker ID *before* asking the compute provider to launch the Worker, give the compute provider that generated ID and use that generated ID to fetch information about the Worker from the Temporal Server once you get a success response from the async call for the compute provider invoking/launching the Worker (function). Sure, there is [`WorkerOptions.Identity`](https://github.com/temporalio/sdk-go/blob/v1.40.0/internal/worker.go#L265) but that isn't an identifier for the Worker, that is more of the User-Agent/client identitity that launched the worker.

wci/workflow/scaling_algorithm/registry.go
02strich: I need a way to group the compute/scaler configs for the different task-type groups together. This is a string version of the task-type-set (computed in a stable fashion). It does work, but not a huge fan so happy for better ideas.

wci/workflow/compute_provider/registry.go
02strich: 1. yes, will change (if the API still exists in the end)
2. the challenge I have with the API overall is that it is sometimes batch create, and sometimes just setting of an integer. This API tries to make them both fit into the same model, but I personally agree on the batch create. That said, I do need an API that allows me to adjust the worker count (up and down) in bigger increments than 1 - otherwise it is a quite inflexible API. I have been toying with the idea of just having 2 APIs: one with a StartWorkerInstance and one with a SetCount or the like command. Thoughts?

There is work on a worker key for something similar, as you are talking about but I am not clear how that would help me here?

wci/workflow/compute_provider/registry.go
```diff
@@ -0,0 +1,75 @@
...
+
+		// ValidateConfig checks the provided configuration for correctness. This might involved
+		// invoking cloud provider functions to check permissions as well. If any issues are found
+		// returns an error with a description
+		ValidateConfig(ctx context.Context, config iface.ComputeProviderConfig) error
+
+		// StartNewWorkerInstance starts a new worker instance when using the LaunchStrategy 'invoke'. any
+		// error is returned if the invocation failed, but not if the invoked work dies before connecting
+		// to Temporal. Returns an error for other launch strategies.
+		StartNewWorkerInstance(ctx context.Context, config iface.ComputeProviderConfig) error
```
jaypipes: Consider renaming to `InvokeWorker`?

jaypipes (commented):
lgtm, have a suggestion for renaming the one API call, but overall looks good.

wci/workflow/compute_provider/registry.go
02strich: yeah that is more consistent - done

jaypipes (approved):

PR #7, opened 2026-03-02
Add a set of compute provider implementations

02strich:
## What was changed
This adds a first set of compute provider implementations for real compute providers AWS, GCP and K8s as well as a test provider for development (subprocess)

## Why?
Compute Providers integrate the auto-scaling decisions with the customers compute environment. A wide-variety allows for customer choice and breadth of coverage.

### DIFF
--- a/wci/client/config.go
+++ b/wci/client/config.go
@@ -2,6 +2,18 @@ package client
 
 import "go.temporal.io/server/common/dynamicconfig"
 
+type (
+	AWSIAMRoleRequest struct {
+		RoleARN         string
+		RoleSessionName string
+		ExternalID      *string
+	}
+
+	GCPIAMServiceAccountRequest struct {
+		ServiceAccountEmail string
+	}
+)
+
 var (
 	WorkerControllerEnabled = dynamicconfig.NewNamespaceBoolSetting(
 		"workercontroller.enabled",
@@ -17,6 +29,16 @@ var (
 		"workercontroller.instanceWorkflowVersion",
 		0,
 		`WorkerControllerInstanceWorkflowVersion controls what version of the logic should the manager workflows use.`,
+	)
+	WorkerControllerAWSIntermediaryRoles = dynamicconfig.NewGlobalTypedSetting(
+		"workercontroller.compute_providers.aws.intermediary_roles",
+		[][]AWSIAMRoleRequest(nil),
+		`WorkerControllerAWSIntermediaryRoles defines the role chain to assume before assuming the per-provider role.`,
+	)
+	WorkerControllerGCPIntermediaryServiceAccounts = dynamicconfig.NewGlobalTypedSetting(
+		"workercontroller.compute_providers.gcp.intermediary_service_accounts",
+		[][]GCPIAMServiceAccountRequest(nil),
+		`WorkerControllerGCPIntermediaryServiceAccounts defines the service account chaining to assume before assuming the per-provider role.`,
 	)
 	WorkerControllerEnabledComputeProviders = dynamicconfig.NewGlobalTypedSetting(
 		"workercontroller.compute_providers.enabled",
--- a/wci/workflow/compute_provider/aws.go
+++ b/wci/workflow/compute_provider/aws.go
@@ -1,0 +1,108 @@
+package computeprovider
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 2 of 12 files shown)


wci/client/config.go
```diff
@@ -18,6 +30,16 @@ var (
 		0,
 		`WorkerControllerInstanceWorkflowVersion controls what version of the logic should the manager workflows use.`,
 	)
+	WorkerControllerAwsIntermediaryRoles = dynamicconfig.NewGlobalTypedSetting(
```
jaypipes: style nit: AWS should be capitalized, as should ARN, ID, ECS and GCP throughout per Go's [initialisms guidance](https://go.dev/wiki/CodeReviewComments#initialisms).

wci/client/config.go
```diff
@@ -18,6 +30,16 @@ var (
...
+		"workercontroller.compute_providers.aws.intermediary_roles",
```
jaypipes, line 34: Couple things...

Why are we supporting multiple roles here? What is the use case for that?

I'm not sure "intermediary_roles" is a great name for this... AWS itself doesn't call it that :) Maybe call it "Service Role" or "Trusted Role" and make the description something like:

> WorkerControllerAWSServiceRole is the ARN of the AWS IAM Role under which the Worker Controller will run. Customers will need to grant this Role permission to calls AWS STS AssumeRole in order to effect changes in the customer's AWS environment.

wci/client/config.go
```diff
@@ -18,6 +30,16 @@ var (
...
+		[][]AWSIAMRoleRequest(nil),
+		`WorkerControllerAwsIntermediaryRoles defines the role chain to assume before assuming the per-provider role.`,
+	)
+	WorkerControllerGcpIntermediaryServiceAccounts = dynamicconfig.NewGlobalTypedSetting(
+		"workercontroller.compute_providers.gcp.intermediary_service_accounts",
+		[][]GCPIAMServiceAccountRequest(nil),
+		`WorkerControllerGcpIntermediaryServiceAccounts defines the service account chaining to assume before assuming the per-provider role.`,
+	)
```
jaypipes: Same question about use case for having multiple service accounts.

I'd recommend naming this `WorkerControllerGCPServiceAccount` and use the following description:

> WorkerControllerGCPServiceAccount is the email address of the GCP Service Account under which the Worker Controller will run. Customers will need to grant this Service Account the Service Account Token Creator IAM Role (`roles/iam.serviceAccountTokenCreator`) in order to effect changes in the customer's GCP environment.

wci/workflow/compute_provider/aws.go
```diff
@@ -0,0 +1,111 @@
...
+		return fmt.Errorf("role ARN must start with 'arn': %s", roleArn)
+	}
+	if parts[2] != "iam" {
+		return fmt.Errorf("role ARN must be an IAM resource: %s", roleArn)
+	}
+	if !strings.HasPrefix(parts[5], "role/") {
+		return fmt.Errorf("role ARN must reference a role: %s", roleArn)
+	}
+	return nil
+}
```
jaypipes: aws-sdk-go-v2 has an `aws/arn.Parse` method that you can use for the above:

https://github.com/aws/aws-sdk-go-v2/blob/e64a7d0f3ff5248deaf30e95b113355b1db3816e/aws/arn/arn.go#L61

wci/client/config.go
02strich: One day I will remember it... Fixed now I think

wci/workflow/compute_provider/aws.go
02strich: Oh thank you - hadn't seen it, but now using it :)

wci/client/config.go
02strich: The setup in tcloud uses role chaining, where you first assume a series of intermediary roles before assuming the customer role. This structure is more-or-less directly replicating the current approach, and there isn't really any material issue with it.

AWS doesn't use this concept b/c they get SLRs and service principals, which you don't get outside.

wci/client/config.go
02strich: Same structure as for AWS above

wci/client/config.go
jaypipes: ack, thanks for the explanation! :)

wci/workflow/compute_provider/aws.go
```diff
@@ -0,0 +1,108 @@
...
+		*result.Credentials.AccessKeyId,
+		*result.Credentials.SecretAccessKey,
+		*result.Credentials.SessionToken,
+	)
+
+	return cfg, nil
+}
+
+func extractRegionFromARN(arnString string) (string, error) {
+	parsedARN, err := arn.Parse(arnString)
```
jaypipes, line 86: muy bueno :)

jaypipes (approved):

PR #8, opened 2026-03-18
Add implementation for no-sync-match algo

02strich:
## What was changed
Adding a scaling algo implementation for function providers, e.g. AWS Lambda. This is focused on the no-sync-match signals with some backstops for error cases.

## Why?
This is one of the last pieces missing to make function providers work, and is likely the simplest meaningful algorithm for these platforms.

### DIFF
--- a/wci/workflow/activities.go
+++ b/wci/workflow/activities.go
@@ -175,6 +175,9 @@ func (a *Activities) UpdateWorkerSetSize(ctx context.Context, req *UpdateWorkerS
 func (a *Activities) HandleTaskAddSignal(ctx context.Context, req HandleTaskAddSignalActivityRequest) (*HandleTaskAddSignalActivityResponse, error) {
 	logger := activity.GetLogger(ctx)
 	updatedScalingStatus := req.ScalingStatus
+	if updatedScalingStatus == nil {
+		updatedScalingStatus = map[string]iface.ScalingAlgorithmStatus{}
+	}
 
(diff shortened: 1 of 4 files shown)


smuneebahmad:
In `HandleTaskAddSignal()` the `updatedScalingStatus` will be `nil` [here](https://github.com/temporalio/temporal-managed-workers/blob/f88b7fd08d86db11857f6b185eb06a79509d2b79/wci/workflow/activities.go#L177) and trigger a panic at [activities.go:204](https://github.com/temporalio/temporal-managed-workers/blob/f88b7fd08d86db11857f6b185eb06a79509d2b79/wci/workflow/activities.go#L204). This happens on the first `HandleTaskAddSignal()` call for a new WCI workflow, when no scaling status is available yet.

02strich:
> In `HandleTaskAddSignal()` the `updatedScalingStatus` will be `nil` [here](https://github.com/temporalio/temporal-managed-workers/blob/f88b7fd08d86db11857f6b185eb06a79509d2b79/wci/workflow/activities.go#L177) and trigger a panic at [activities.go:204](https://github.com/temporalio/temporal-managed-workers/blob/f88b7fd08d86db11857f6b185eb06a79509d2b79/wci/workflow/activities.go#L204). This happens on the first `HandleTaskAddSignal()` call for a new WCI workflow, when no scaling status is available yet.

good catch, fixed

smuneebahmad (approved):

PR #10, opened 2026-03-18
Add ValidateWorkerControllerInstanceSpec API

02strich:
## What was changed
Adding a new API to validate an instance spec without applying it anywhere. This allows customers to get dry-run feedback on their configuration.

## Why?
Users might want to check their config before committing it - this gives them feedback.

### DIFF
--- a/wci/client/workflow_interaction.go
+++ b/wci/client/workflow_interaction.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"errors"
+	"time"
 
 	"github.com/temporalio/temporal-managed-workers/wci/workflow/iface"
 	commonpb "go.temporal.io/api/common/v1"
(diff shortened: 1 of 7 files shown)


wci/client/workflow_interaction.go
```diff
@@ -20,8 +21,108 @@ import (
...
+				return serviceerror.NewInternal(failure.GetMessage())
+			}
+			if event.GetWorkflowExecutionTimedOutEventAttributes() != nil {
+				return serviceerror.NewInternal("workflow execution timed out")
+			}
+			if event.GetWorkflowExecutionCanceledEventAttributes() != nil {
+				return serviceerror.NewInternal("workflow execution was cancelled")
+			}
+			if event.GetWorkflowExecutionTerminatedEventAttributes() != nil {
+				return serviceerror.NewInternal("workflow execution was terminated")
```
jaypipes: consider making the error message include the workflow name/type, otherwise these errors may be hard to distinguish.

wci/workflow/validate_workflow.go
```diff
@@ -0,0 +1,37 @@
...
+		return temporal.NewApplicationError("spec must be provided", "InvalidArgument")
+	}
+
+	if err := args.Spec.Validate(); err != nil {
+		return temporal.NewApplicationError(err.Error(), "InvalidArgument")
+	}
+
+	err := workflow.ExecuteActivity(
+		workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+			StartToCloseTimeout: 2 * time.Second,
```
jaypipes: might be good to define this as a package-scoped var instead of a hard-coded constant.

wci/workflow/validate_workflow.go
```diff
@@ -0,0 +1,37 @@
...
+			RetryPolicy:         &temporal.RetryPolicy{MaximumAttempts: 1},
```
jaypipes, line 23: ditto.

jaypipes (approved):
minor comments, code looks good to me!

wci/client/workflow_interaction.go
02strich: Added the request ID as that is the most innocent to return to customers (in an edge case) and is still searchable. Good idea!

wci/workflow/validate_workflow.go
02strich: yeah, did it across the board now - thanks!

wci/workflow/validate_workflow.go
02strich: ditto

PR #12, opened 2026-03-19
Check whether the external ID is actually used during config validation

02strich:
## What was changed
There is a risk of a customer specifying an external ID with us, but not actually using it correctly in their policy. This adds a check as part of the config validation.

## Why?
Adds validation that the external ID is correctly setup if provided to avoid a false sense of security

### DIFF
--- a/wci/workflow/compute_provider/aws.go
+++ b/wci/workflow/compute_provider/aws.go
@@ -2,6 +2,7 @@ package computeprovider
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"strings"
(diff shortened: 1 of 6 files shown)


wci/workflow/compute_provider/aws.go
jaypipes, line 43: Honestly, I still have issues with this. I know this is copied code from a different temporal repo (saas-control-plane maybe?) but the whole "select a random role ARN from an array of intermediary roles" is just weird, IMHO. Especially so considering there is no code comment indicating why this is happening, what the side-effects of this non-determinism might be, and what goal this is accomplishing.

wci/workflow/compute_provider/aws.go
```diff
@@ -51,6 +62,42 @@ func buildAWSConfig(ctx context.Context, region, roleARN string, externalID *str
...
+	})
+	if err == nil {
+		return fmt.Errorf("role %s does not require the configured external ID; update the role's trust policy to enforce the aws:ExternalId condition", roleARN)
+	}
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) && apiErr.ErrorCode() == "AccessDenied" {
+		return nil
+	}
+	return fmt.Errorf("unable to verify external ID enforcement for role %s: %w", roleARN, err)
+}
```
jaypipes, line 93: Did an AI coding agent generate this code and the unit tests? Because the unit tests will always pass because the mocks essentially enforce the contract from this unit of code instead of enforcing the contract from the upstream aws-sdk-go-v2 library. The upstream aws-sdk-go-v2 library has a `types.AccessDeniedException` error that can be type-compared like so:

```go
        var apiError types.AccessDeniedException
        if errors.As(err, &apiError) {
            // it's a 403 Access Denied...
        }
...
```

wci/workflow/compute_provider/aws.go
02strich: I added comments to at least explain it now. It is a thing I am working in a bigger theme to revisit.

wci/workflow/compute_provider/aws.go
02strich: as discussed, STS doesn't have `AccessDeniedException` so the current code is accurate

jaypipes (approved):

PR #16, opened 2026-03-20
Enforce role and external ID in AWS compute providers

02strich:
## What was changed
Add enforcement of the external ID and associated confused deputy protection. Also offers opt-out for special cases.

## Why?
Confused Deputy protection is an important security safeguard, that also self-hosted customers should consider. Enforcing it
by default provides a good incentive to make it work - and opt-out is still available for other cases.


smuneebahmad (approved):

jaypipes (approved):
yup, makes sense to me!

PR #18, opened 2026-03-25
Update repo name to be more specific

02strich:
## What was changed
Updating the repo name to be more specific to the scope

## Why?
The original name was too generic and ambigious, raising questions in the future. The new name is more specific.


cmd/worker/main.go
smuneebahmad, line 1: Should we also update `Usage:` on lines: 30 and 54?

02strich: good catch. done

smuneebahmad (approved):

PR #20, opened 2026-03-26
Update Workflow to new API

02strich:
## What was changed
Updating the spec to the adjusted data structures in the public API (using an explicit ID, the name scaling group and the removal of the explicit task queue in the API)

## Why?
The external API has recently changed slightly based on learnings from the project. This brings along the internal API to avoid friction.


jaypipes (approved):
Good stuff, @02strich :)

PR #27, opened 2026-03-31
Bring Payload and Incremental Update handling into WCI

02strich:
## What was changed
* Replace map[string]any in WCI with a Payload
* Move incremental update processing into the WCI workflow

## Why?
* Eventually the WCI needs to handle the Payload so brining that in now to avoid later interface changes
* Doing the incremental update processing in the Worker Deployment workflow means there is a risk of doing it non-transactionally. It is easier to do it directly in the WCI workflow

### DIFF
--- a/wci/client/client.go
+++ b/wci/client/client.go
@@ -37,6 +37,10 @@ const (
 
 type (
 	Client interface {
+		UpdateTaskQueueName(
+			ctx context.Context,
+			newTaskQueueName string,
+		) error
 		WorkerControllerInstanceExists(
 			ctx context.Context,
 			namespaceEntry *namespace.Namespace,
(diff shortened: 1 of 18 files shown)


smuneebahmad (approved):

wci/client/client.go
ShahabT: can you make this also incremental so we can get rid of the merge utils completely from temporal?

ShahabT: right now I describe the instance to get the config, then apply merge logic then call this method. it seems you validation workflow can do the describe and merge steps itself.

02strich: Yeah had forgotten you changed the validate api to support incremental

smuneebahmad (approved):

ShahabT (approved):

PR #30, opened 2026-04-01
Fix issue with waiting for workflow completion

02strich:
## What was changed
In the temporal server tests the response from HistoryService is slightly different than when it runs normally. To be compatible with both cases, this adds a little bit more flexibility.

### DIFF
--- a/wci/client/workflow_interaction.go
+++ b/wci/client/workflow_interaction.go
@@ -91,9 +91,11 @@ func startWorkflowAndWait(
 			return err
 		}
 
-		events := resp.GetResponse().GetHistory().GetEvents()
+		events := resp.GetHistory().GetEvents()
 		if len(events) > 0 {
-			event := events[0]
+			// the very last event is the one we are looking for
+			event := events[len(events)-1]
+
 			if event.GetWorkflowExecutionCompletedEventAttributes() != nil {
 				return nil
 			}


wci/client/workflow_interaction.go
```diff
@@ -91,9 +91,12 @@ func startWorkflowAndWait(
 			return err
 		}
 
-		events := resp.GetResponse().GetHistory().GetEvents()
-		if len(events) > 0 {
-			event := events[0]
+		events := resp.GetHistory().GetEvents()
+		for _, event := range events {
+			if event.EventType == enumspb.EVENT_TYPE_WORKFLOW_TASK_COMPLETED {
```
ShahabT: I think we should skip any event type other than the few you check below.

wci/client/workflow_interaction.go
ShahabT, line 121: Not sure if this makes sense... if the event type doesn't match one of the ones above we should continue the loop.

wci/client/workflow_interaction.go
02strich: The API is expected to already filter out all other events

wci/client/workflow_interaction.go
02strich: The HistoryService API call is already filtering the event list so that all events returned are about the workflow closing

wci/client/workflow_interaction.go
ShahabT: oh ok.

ShahabT (approved):

smuneebahmad (approved):

PR #35, opened 2026-04-07
Return the updated spec on update

02strich:
## What was changed
Returning the updated spec on update to avoid having to make another call to get it.

## Why?
The Worker Deployment Version keeps track of summary-level details on the spec. Having the updated spec returned simplifies that bookkeeping.


jaypipes (approved):
👍 muy bueno.

smuneebahmad (approved):

ShahabT (approved):

PR #42, opened 2026-04-17
Add automatic validation of Compute Config

02strich:
## What was changed
Adds a tracking of the validation state of the compute config. When an update is successful, that means the spec is good. The spec is re-checked on a cadence to make sure it still works. The validation state is exposed via the query for upstream to expose it to external callers.

## Why?
A compute config that was valid when it got stored, might become invalid by virtue of changes outside of the temporal server (e.g. the AWS IAM config changing). This adds regular validation so we can highlight that to customers.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/wci/workflow/workflow.go
+++ b/wci/workflow/workflow.go
@@ -134,6 +134,7 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 			ConflictToken:        d.State.ConflictToken,
 			CreateTime:           d.State.CreateTime,
 			LastModifierIdentity: d.State.LastModifierIdentity,
+			ValidationState:      d.State.ValidationState,
 		}, nil
 	}); err != nil {
 		return err
@@ -270,7 +271,14 @@ func (d *WorkflowRunner) handleUpdateInstance(ctx workflow.Context, args *iface.
 			return &iface.UpdateWorkerControllerInstanceResponse{Spec: d.State.Spec}, nil
 		}
 
+		validationTime := workflow.Now(ctx)
+
 		if err := updatedSpec.Validate(); err != nil {
+			d.State.ValidationState = &iface.ValidationState{
+				LastValidatedTime: timestamppb.New(validationTime),
+				Status:            iface.ValidationStatusFailed,
+				ErrMessage:        err.Error(),
+			}
 			return nil, serviceerror.NewInvalidArgumentf("%w", err)
 		}
 
@@ -280,11 +288,19 @@ func (d *WorkflowRunner) handleUpdateInstance(ctx workflow.Context, args *iface.
 			&ValidateSpecRequest{Spec: updatedSpec},
 		).Get(ctx, nil); err != nil {
 			var appErr *temporal.ApplicationError
+			errMsg := err.Error()
+			if errors.As(err, &appErr) {
+				errMsg = appErr.Message()
+			}
+			d.State.ValidationState = &iface.ValidationState{
+				LastValidatedTime: timestamppb.New(validationTime),
+				Status:            iface.ValidationStatusFailed,
+				ErrMessage:        errMsg,
+			}
 			if errors.As(err, &appErr) {
 				return nil, serviceerror.NewInvalidArgumentf("%s", appErr.Message())
-			} else {
-				return nil, err
 			}
+			return nil, err
 		}
 
 		// we need to scale up each of the groups for a moment to get them to register the task queues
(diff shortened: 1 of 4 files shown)


wci/workflow/workflow.go
```diff
@@ -201,21 +208,20 @@ func (d *WorkflowRunner) handleValidateSpec(ctx workflow.Context, args *iface.Va
 		RemoveScalingGroups: args.RemoveScalingGroups,
 	})
 	if err != nil {
-		return nil, serviceerror.NewInvalidArgumentf("%w", err)
+		return nil, serviceerror.NewInvalidArgumentf("%s", err.Error())
```
jaypipes, line 211: Here and below, why the changes to the error handling in this method? Seems orthogonal to the purpose of the PR...

wci/workflow/workflow.go
```diff
@@ -382,6 +386,35 @@ func (d *WorkflowRunner) pullStatsAndUpdate(ctx workflow.Context) time.Duration
...
+		}),
+		d.a.ValidateSpec,
+		&ValidateSpecRequest{Spec: d.State.Spec},
+	).Get(ctx, nil); err != nil {
+		if appErr, ok := errors.AsType[*temporal.ApplicationError](err); ok {
+			d.State.ValidationStatus = iface.NewValidationStatusFailed(now, appErr.Message())
+			d.logger.Warn("Periodic spec validation failed with spec error", "error", err)
+		} else {
+			// Transient infrastructure errors (timeouts, server errors, cancellation) are not
+			// spec failures — leave ValidationStatus at its last known value.
```
jaypipes, line 408: That emdash in the above code comment... hi Claude! :)

wci/workflow/workflow.go
```diff
@@ -134,6 +140,7 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 			ConflictToken:        d.State.ConflictToken,
 			CreateTime:           d.State.CreateTime,
 			LastModifierIdentity: d.State.LastModifierIdentity,
+			ValidationStatus:     d.State.ValidationStatus,
```
jaypipes, line 143: This is a great example of what I dislike about the Temporal Workflow data model and why using (entity) Temporal Workflows to implement everything inside a daemon like WCI leads to confusing state management.

The "validation status" should not be a state field on the Workflow itself. It should be a state field on the **WorkerDeploymentVersion** object because it represents whether the compute config associated with that WorkerDeploymentVersion is valid. But because use of the entity workflow pattern inside Temporal Server and its sub-services is so prevalent, this is really the only way to do this kind of thing. 

The main user-facing problem that manifests from this architecture is that the user does not see the state of these internal entity workflows and so instead of just updating the WorkerDeploymentVersion.ComputeConfig field atomically with the current validation status, we will now have to add more code into the frontend server to pull information from the WCI workflow's state out into the returned response to a DescribeWorkerDeploymentVersion API call :( Which is, frankly, an over-engineered Rube Goldberg-esque process that should be a simple single atomic database write.

wci/workflow/workflow.go
```diff
@@ -382,6 +386,35 @@ func (d *WorkflowRunner) pullStatsAndUpdate(ctx workflow.Context) time.Duration
...
+			// Transient infrastructure errors (timeouts, server errors, cancellation) are not
+			// spec failures — leave ValidationStatus at its last known value.
+			d.logger.Warn("Periodic spec validation failed with transient error, leaving validation state unchanged", "error", err)
+		}
+	} else {
+		d.State.ValidationStatus = iface.NewValidationStatusSuccess(now)
+	}
+
+	// We are not setting stateChanged to true to avoid unneccessary CaNs here.
+}
```
jaypipes, line 416: to add to my comment above about the ramifications of using the entity workflow pattern everywhere, if we weren't using entity workflows and were instead able to simply atomically update the WorkerDeploymentVersion object directly in a single API call, this entire method could be written in an easier-to-reason-about way that is more explicit and less magical.

```go
func (d *WorkflowRunner) periodicValidateSpec(ctx context.Context) {
	if d.State == nil || d.State.Spec == nil || len(d.State.Spec.ScalingGroupSpecs) == 0 {
		return
	}
	existingValid := d.ValidationStatus
	newValid := d.validateSpec(...) // would not need to be an activity...
	if existingValid != newValid {
		return d.updateWorkerDeploymentVersionComputeConfigValidateStatus(newValid)
	}
}
```

and the above method could be run in a standard Go time.Ticker loop instead of inside an entity workflow. You wouldn't need to be concerned about CaNs or any of the other stuff that goes along with temporal workflows for such a basic "update this object if state has changed" use case.

jaypipes (approved):
Works for the existing codebase. I've left my requisite diatribe about entity workflows inline. :)

wci/workflow/workflow.go
02strich: Lint found it while I was working on this. No idea how I missed it originally. Didn't seem worthwhile to split out into its own pr

PR #46, opened 2026-04-28
Feat: WCI workflow health metrics

zainawaisn:
## What was changed
Adds a wci level `metrics` package. This contains the initial set of observability metrics specifically for WCI workflow health. 

## Why?
Initial observability for serverless workers pre-release. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/wci/metrics/metric_defs.go
+++ b/wci/metrics/metric_defs.go
@@ -3,27 +3,49 @@ package metrics
 import "go.temporal.io/server/common/metrics"
 
 var (
-	WorkerControllerInstancePullStatsCount       = metrics.NewCounterDef("worker_controller_instance_pull_stats_count")
-	WorkerControllerInstancePullStatsErrorCount  = metrics.NewCounterDef("worker_controller_instance_pull_stats_error_count")
-	WorkerControllerInstanceBacklogDetectedCount = metrics.NewCounterDef("worker_controller_instance_backlog_detected_count")
+	WorkerControllerInstanceBacklogDetectedCount = metrics.NewCounterDef(
+		"worker_controller_instance_backlog_detected_count",
+		metrics.WithDescription("The number of times a backlog was detected for a worker controller instance."))
 
-	WorkerControllerInstanceScaleUpCount          = metrics.NewCounterDef("worker_controller_instance_scale_up_count")
-	WorkerControllerInstanceScaleUpThrottledCount = metrics.NewCounterDef("worker_controller_instance_scale_up_throttled_count")
+	WorkerControllerInstanceScaleUpCount          = metrics.NewCounterDef(
+		"worker_controller_instance_scale_up_count",
+		metrics.WithDescription("The number of times a scale up was performed for a worker controller instance."))
+	WorkerControllerInstanceScaleUpThrottledCount = metrics.NewCounterDef(
+		"worker_controller_instance_scale_up_throttled_count",
+		metrics.WithDescription("The number of times a scale up was throttled for a worker controller instance."))
 
-	WorkerControllerInstanceInvokeWorkerCount      = metrics.NewCounterDef("worker_controller_instance_invoke_worker_count")
-	WorkerControllerInstanceInvokeWorkerErrorCount = metrics.NewCounterDef("worker_controller_instance_invoke_worker_error_count")
+	WorkerControllerInstanceWorkflowErrorCount = metrics.NewCounterDef(
+		"worker_controller_instance_workflow_error_count",
+		metrics.WithDescription("The number of times a workflow error occurred for a worker controller instance."))
 
-	WorkerControllerInstanceWorkflowErrorCount = metrics.NewCounterDef("worker_controller_instance_workflow_error_count")
-	WorkerControllerInstanceTaskAddSignalCount = metrics.NewCounterDef("worker_controller_instance_task_add_signal_count")
+	WorkerControllerInstanceSignals = metrics.NewCounterDef(
+		"worker_controller_instance_signals",
+		metrics.WithDescription("The number of times a signal occurred for a worker controller instance."))
+
+	WorkerControllerInstanceOperations = metrics.NewCounterDef(
+		"worker_controller_instance_operations",
+		metrics.WithDescription("The number of times an operation occurred for a worker controller instance."))
 )
 
 // Tag key constants matching go.temporal.io/server/common/metrics/tags.go.
+// Exported tag constructors do not work in SDK metrics handlers so tags are defined as constants here.
 // Used as map keys with the SDK MetricsHandler (workflow and activity context).
 const (
 	NamespaceTag               = "namespace"
 	ActivityTypeTag            = "activityType"
 	WorkerDeploymentNameTag    = "worker_deployment_name"
 	WorkerDeploymentBuildIDTag = "worker_build_id"
+	SignalTypeTagName          = "signal_type"
+	OperationTagName           = metrics.OperationTagName
+	ErrorTypeTagName           = metrics.ErrorTypeTagName
+
 	// TriggerTag is for future usage in scale up and throttling metrics
-	// TriggerTag                 = "trigger"
+	// TriggerTag              = "trigger"
+)
+
+// Tag value constants
+const (
+	SignalTypeTaskAdd = "task_add"
+	OperationTypePullStats = "pull_stats"
+	OperationTypeInvokeWorker = "invoke_worker"
 )
--- a/wci/metrics/metric_defs.go
+++ b/wci/metrics/metric_defs.go
@@ -7,7 +7,7 @@ var (
 		"worker_controller_instance_backlog_count",
 		metrics.WithDescription("The total detected backlog size for a worker controller instance."))
 
-	ScaleUpCount          = metrics.NewCounterDef(
+	ScaleUpCount = metrics.NewCounterDef(
 		"worker_controller_instance_scale_up_count",
 		metrics.WithDescription("The number of times a scale up was performed for a worker controller instance."))
 	ScaleUpThrottledCount = metrics.NewCounterDef(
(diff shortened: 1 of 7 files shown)


ericbriffa:
Should we add units to ensure the values are emitted as we expect? I can think of two reasons why: 
1. The slow feedback loop of discovering they are not as expected and having to cut another PR and merge. 
2. Metrics regressions can bite us extra hard as we use these to monitor the state of the service itself.

jaypipes (changes requested):
Hold off on merging this please. I have some concerns about the naming and structure of these metrics.

wci/metrics/metric_defs.go
```diff
@@ -0,0 +1,29 @@
...
+
+	WorkerControllerInstanceInvokeWorkerCount      = metrics.NewCounterDef("worker_controller_instance_invoke_worker_count")
+	WorkerControllerInstanceInvokeWorkerErrorCount = metrics.NewCounterDef("worker_controller_instance_invoke_worker_error_count")
+
+	WorkerControllerInstanceWorkflowErrorCount = metrics.NewCounterDef("worker_controller_instance_workflow_error_count")
+	WorkerControllerInstanceTaskAddSignalCount = metrics.NewCounterDef("worker_controller_instance_task_add_signal_count")
+)
+
+// Tag key constants matching go.temporal.io/server/common/metrics/tags.go.
+// Used as map keys with the SDK MetricsHandler (workflow and activity context).
```
02strich, line 32: can you add a note to explain why they needed to be replicated here instead of using them from the server?

wci/workflow/scaling_algorithm/registry.go
```diff
@@ -35,6 +35,7 @@ type (
 		Actions []ScalingAction
 		// The updated scaling status to be persisted until the next run
 		Status iface.ScalingAlgorithmStatus
+		ThrottledCount int
```
02strich, line 39: please add a comment explaining what this is

wci/metrics/metric_defs.go
zainawaisn: tags themselves are unexported in the server. to use tags externally, you have to use the exported constructors that are provided, but the sdk metrics handler cannot take in the constructers as an argument to MetricsHandler.WithTags(). Will add a note.

jaypipes (changes requested):
OK, summarizing feedback...

1. Combine the following Counters:

* `worker_controller_instance_pull_stats_count`
* `worker_controller_instance_pull_stats_error_count`
* `worker_controller_instance_invoke_worker_count`
* `worker_controller_instance_invoke_worker_error_count`

into a single Counter called `worker_controller_instance_operations`. The OperationType and ErrorType tags will be used on this single Counter.

2. Rename the `worker_controller_instance_task_add_signal_count` Counter to `worker_controller_instance_signals`. Add a `SignalTypeTagName` constant with a value of `signal_type` and a `SignalTypeTaskAdd` constant with a value of `task_add`.

3. Add Descriptions for all Counters.

wci/workflow/activities.go
```diff
@@ -338,6 +352,18 @@ func (a *Activities) PullStats(ctx context.Context, req *PullStatsActivityReques
 
 	logger.Info("Pull Stats Results", "workflow_count", metricsSnapshot.Workflow.LastBacklogCount, "activity_count", metricsSnapshot.Activity.LastBacklogCount, "nexus_count", metricsSnapshot.Nexus.LastBacklogCount)
 
+	totalBacklog := metricsSnapshot.Workflow.LastBacklogCount + 
+					metricsSnapshot.Activity.LastBacklogCount +
+					metricsSnapshot.Nexus.LastBacklogCount
+
```
zainawaisn, line 358: Open question - Should we emit when backlog is 0? This would technically remove ambiguity whether PullStats is working or not, but we already have the `Operations` metric with `OperationTypePullStats` that tracks this.

wci/metrics/metric_defs.go
```diff
@@ -0,0 +1,51 @@
+package metrics
+
+import "go.temporal.io/server/common/metrics"
+
+var (
+	WorkerControllerInstanceBacklogCount = metrics.NewGaugeDef(
```
02strich: I thought we were going to shorten the var names?

wci/metrics/metric_defs.go
```diff
@@ -0,0 +1,51 @@
...
+		"worker_controller_instance_backlog_count",
+		metrics.WithDescription("The total detected backlog size for a worker controller instance."))
+
+	WorkerControllerInstanceScaleUpCount          = metrics.NewCounterDef(
```
02strich: formatting with the spaces seems off here

wci/metrics/metric_defs.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// TriggerTag is for future usage in scale up and throttling metrics
+	// TriggerTag              = "trigger"
+)
+
+// Tag value constants
+const (
+	SignalTypeTaskAdd = "task_add"
+	OperationTypePullStats = "pull_stats"
+	OperationTypeInvokeWorker = "invoke_worker"
```
jaypipes: you need to run `go fmt -w` on code before submitting it, as it will properly  format the Go code above to align things like block above it for Tag names

wci/metrics/metric_defs.go
02strich: also isn't scale_up an operation?

wci/metrics/metric_defs.go
```diff
@@ -0,0 +1,51 @@
...
+		"worker_controller_instance_operations",
+		metrics.WithDescription("The number of times an operation occurred for a worker controller instance."))
+)
+
+// Tag key constants matching go.temporal.io/server/common/metrics/tags.go.
+// Exported tag constructors do not work in SDK metrics handlers so tags are defined as constants here.
+// Used as map keys with the SDK MetricsHandler (workflow and activity context).
+const (
+	NamespaceTag               = "namespace"
+	ActivityTypeTag            = "activityType"
```
02strich, line 35: why not snake_case like the others?

wci/workflow/activities.go
02strich: I personally prefer constant emission b/c there are too many systems who garbage collect metrics otherwise

wci/workflow/scaling_algorithm/no_sync_match.go
```diff
@@ -147,22 +147,23 @@ func (a *scalingAlgorithmNoSync) ProcessTaskAdd(ctx context.Context, config ifac
...
 		nowMs := time.Now().UnixMilli() // safe: called from activity context, not workflow
 		elapsedMs := nowMs - lastScaleUpMs
 
-		// TODO: add metric for throttling
 		if elapsedMs >= cooloffMs {
 			actions = append(actions, ScalingAction{Action: ActionTypeInvokeWorker})
 			updatedState[stateLastScaleUpTimestampKey] = nowMs
 		} else {
 			logger.Info("Throttled worker invocation", "elapsed_ms", elapsedMs)
+			throttledCount = 1
```
jaypipes: Seems to me that `TaskAddReponse.ThrottledCount` isn't actually a count of anything but rather it's just a boolean of whether throttling occurred. Is that correct?

wci/workflow/activities.go
```diff
@@ -274,6 +279,15 @@ func (a *Activities) HandleTaskAddSignal(ctx context.Context, req HandleTaskAddS
 			updatedActions = append(updatedActions, act)
 		}
 
+		if response.ThrottledCount > 0 {
+			metricsHandler := activity.GetMetricsHandler(ctx).WithTags(map[string]string{
+				wcimetrics.NamespaceTag:               req.NamespaceName,
+				wcimetrics.WorkerDeploymentNameTag:    req.DeploymentName,
+				wcimetrics.WorkerDeploymentBuildIDTag: req.DeploymentBuildID,
+			})
+			metricsHandler.Counter(wcimetrics.WorkerControllerInstanceScaleUpThrottledCount.Name()).Inc(int64(response.ThrottledCount))
```
jaypipes: Is ThrottledCount ever not either 1 or 0?

jaypipes (commented):
few small comments..

wci/metrics/metric_defs.go
zainawaisn: keeping the tag exactly how temporal server does it.

02strich: should the other tags then follow that format? I find it odd to have two different ones...

wci/metrics/metric_defs.go
zainawaisn: Done.

wci/metrics/metric_defs.go
zainawaisn: temporal server tags are inconsistent themselves.

wci/workflow/activities.go
zainawaisn: now emitting BacklogCount every PullStats cycle even when backlog is 0.

wci/workflow/scaling_algorithm/no_sync_match.go
zainawaisn: this was a bug, fixed it now. throttledCount now properly carries the the amount of task scale ups that are throttled.

wci/workflow/activities.go
zainawaisn: ThrottledCount carries the the amount of task scale ups that were throttled.

wci/metrics/metric_defs.go
jaypipes: yeah, we discussed this in standup yesterday. we need to re-use the existing constants from temporal server which are already unfortunately inconsistent. :(

jaypipes (approved):
Thanks @zainawaisn, this looks much better :)

02strich (approved):

PR #49, opened 2026-05-05
feat: add gcp-cloud-run-invoke compute provider with OIDC

brucearctor:
## Summary

Adds a new `gcp-cloud-run-invoke` compute provider that uses `LaunchStrategyInvoke` to trigger Cloud Run services via HTTP POST with OIDC identity tokens. This is the GCP analog to the existing `aws-lambda` provider.

## Authentication

Supports two authentication modes:

- **Workload Identity (default)**: When no `service_account` is configured, uses `idtoken.NewTokenSource` which transparently leverages the host's default credentials (GKE Workload Identity, attached SA on Compute Engine/Cloud Run).
- **SA Impersonation**: When a `service_account` is configured, uses `impersonate.IDTokenSource` with the intermediary delegate chain for cross-project or Temporal Cloud scenarios.

Both paths produce OIDC identity tokens with the Cloud Run service URL as the audience, which is what Cloud Run's IAM invoker check requires.

## Relationship to existing `gcp-cloud-run` provider

The existing `gcp-cloud-run` provider uses `LaunchStrategyWorkerSet` to manage Cloud Run WorkerPool instance counts. This new provider complements it by enabling per-task invocation semantics, similar to how `aws-lambda` works.

## Config

```yaml
provider_type: gcp-cloud-run-invoke
# Config details (passed as Payload):
# url: https://my-worker-abc123-uc.a.run.app   (required)
# service_account: invoker@proj.iam.gserviceaccount.com  (optional, omit for Workload Identity)
```

### DIFF
--- a/wci/workflow/compute_provider/gcp_cloudrun_invoke.go
+++ b/wci/workflow/compute_provider/gcp_cloudrun_invoke.go
@@ -1,0 +1,158 @@
+package computeprovider
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"math/rand"
+	"net/http"
+	"net/url"
+	"time"
+
+	"go.temporal.io/auto-scaled-workers/wci/client"
+	"go.temporal.io/auto-scaled-workers/wci/workflow/iface"
+	"go.temporal.io/server/common/dynamicconfig"
+	"google.golang.org/api/idtoken"
+	"google.golang.org/api/impersonate"
+
+	"golang.org/x/oauth2"
+)
+
+const (
+	configGCPCloudRunInvokeURL            = "url"
+	configGCPCloudRunInvokeServiceAccount = "service_account"
+)
+
+type gcpCloudRunInvokeComputeProvider struct {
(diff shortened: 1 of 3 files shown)


brucearctor:
Still a work in progress ... but again, sketch.

OIDC and workload identity desired [ rather than SA key ].  BUT, how to acknowledge source?

If Cloud 'on' GCP --> https://docs.cloud.google.com/iam/docs/workload-identity-federation something like that.  I am happy to configure workload identity on my side, or also things like https://docs.cloud.google.com/iam/docs/workload-identity-federation-with-kubernetes

I can dig in more.  Assuming collaboration desired.  

Or, at least early tester.

brucearctor:
@02strich ! :-p 

We just spoke on the street leaving/going to replay as well.

I do not expect this to be used -- I opened a PR to start another method for communicating.  Even better to meet in person, and setup email/etc.  I'll happily close this, and take things to email, any helpful design docs, and getting wider collab going.

PR #51, opened 2026-05-18
Buffer pending task-add signals across continue-as-new

02strich:
## What was changed
Drain unprocessed task-add signals into local state before continue-as-new and replay them on the next run so signals are not lost. Inline the signal listener into the main goroutine, add a safe activity logger fallback, and cover the new behavior with workflow tests.

## Why?
The workflow crashed with history-too-large in case of constant signal load (as it would not not CaN).

### DIFF
--- a/wci/workflow/iface/workflow.go
+++ b/wci/workflow/iface/workflow.go
@@ -83,6 +83,8 @@ type (
 		// ScalingStatus contains the state information keyd by the ScalingGroups key
 		ScalingStatus map[string]ScalingAlgorithmStatus `json:"scaling_state"`
 
+		PendingTaskAddSignals []*SignalTaskAddRequest `json:"pending_task_add_signals,omitempty"`
+
 		ConflictToken        []byte                 `json:"conflict_token,omitempty"`
 		CreateTime           *timestamppb.Timestamp `json:"create_time,omitempty"`
 		LastModifierIdentity string                 `json:"last_modifier_identity,omitempty"`
--- a/wci/workflow/scaling_algorithm/safe_activity_logger.go
+++ b/wci/workflow/scaling_algorithm/safe_activity_logger.go
@@ -1,0 +1,27 @@
+package scalingalgorithm
+
+import (
+	"context"
+
+	"go.temporal.io/sdk/activity"
+	sdklog "go.temporal.io/sdk/log"
+)
+
+type (
+	noopLogger struct{}
+)
+
+func safeActivityLogger(ctx context.Context) (logger sdklog.Logger) {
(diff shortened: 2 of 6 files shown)


wci/workflow/scaling_algorithm/safe_activity_logger.go
```diff
@@ -0,0 +1,27 @@
...
+
+func safeActivityLogger(ctx context.Context) (logger sdklog.Logger) {
+	logger = noopLogger{}
+	defer func() {
+		if recover() != nil {
+			logger = noopLogger{}
+		}
+	}()
+	return activity.GetLogger(ctx)
+}
```
jaypipes, line 22: what is this? :)

jaypipes (commented):
Is there a Jira issue or some other place that links a description of what problem/incident/root cause this PR is solving? I'd love to read more about it.

wci/workflow/scaling_algorithm/safe_activity_logger.go
02strich: There was a bug that a whole set of the unit tests didn't work b/c in unit test you don't have the activity context. This allows getting the real logger when it is available, or a noopLogger if it is not (i.e. in testing).

ericbriffa:
@jaypipes this PR addressses this bug: https://temporalio.atlassian.net/browse/COM-91

wci/workflow/iface/workflow.go
```diff
@@ -78,6 +78,8 @@ type (
 		// ScalingStatus contains the state information keyd by the ScalingGroups key
 		ScalingStatus map[string]ScalingAlgorithmStatus `json:"scaling_state"`
 
+		PendingTaskAddSignals []*SignalTaskAddRequest `json:"pending_task_add_signals,omitempty"`
```
gcristea-temporal, line 86: Can this be a map where requests get rolled-up (per queue, type, etc)?

gcristea-temporal (approved):
lgtm 👍 

One question wrt storing the pending requests.

wci/workflow/iface/workflow.go
02strich: What would the value of having this as a map?

gcristea-temporal: It would (potentially) use less memory.

02strich: Given that these are just the carry-overs, and the map would require a different processing path, I don't think that is really worth it.

gcristea-temporal: Fair enough, thanks.

PR #52, opened 2026-05-26
Arm periodic stats pull timer on workflow start

02strich:
The stats pull timer was defined but never initially scheduled, so the periodic stats poll would not fire after a fresh workflow run or continue-as-new. Add the initial addStatsPullTimer(maxPollInterval) call to kick off the polling loop.

This was broken in the rebase for the prior PR — oopsie.


gcristea-temporal (approved):
lgtm 👍

PR #53, opened 2026-05-26
Rework function simulator to model no-sync scaling algorithm

02strich:
## What was changed
Add a SPEC.md that describes the no-sync scaling algorithm as implemented and align the simulator with the no-sync algorithm spec: replace consumer lifespan, creation throttle, and failure injection with scale-up cooloff, metrics poll interval, backlog threshold, max worker lifetime, and dispatch-rate epsilon. Split decisions into task-add and metrics-poll paths sharing a single scale-up timer, and add a sim-only worker exit guard that requeues active tasks.

## Why?
The spec is the summary of the algorithm and clearer than either the real or the simulator implementation. As such it is a cleaner starting point for learning what happens. It is also the authoritative source, to which all implementations are compared.

### DIFF
--- a/docs/simulators/function/SPEC.md
+++ b/docs/simulators/function/SPEC.md
@@ -1,0 +1,232 @@
+# No-Sync Scaling Algorithm Spec
+
+This document specifies the behavior of `scalingAlgorithmNoSync`, the `no-sync`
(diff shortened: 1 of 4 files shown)


jaypipes (approved):
I like it. I like it a LOT.

PR #54, opened 2026-05-28
Add EWMA queue consumer autoscaling simulator

02strich:
Adds a browser-based discrete-event simulator under docs/simulators/service-ewma/ that models the EWMA-based queue consumer scaling algorithm, along with a SPEC.md documenting the algorithm's inputs, state, capacity estimate, no-sync decision, and cadence decision logic.

This EWMA algorithm it so far the best candidate for scaling worker set compute providers.

### DIFF
--- a/docs/simulators/workerset-ewma/SPEC.md
+++ b/docs/simulators/workerset-ewma/SPEC.md
@@ -1,0 +1,219 @@
+# Worker Set Scaling Algorithm Spec
+
+This document specifies a worker set scaling algorithm using exponential weighted-moving-average (EWMA) on the arrival/dispatch rates reported by matching service. It determines the size of the worker set (a.k.a. the replica count) with a combination of a steady state, queueing theory-based formula and a fast-path handler for no-sync matches (that quickly triggers upscaling).
+
+The fast path is driven by the clarity of the signal: when a no-sync match happens, it directly means that there is not sufficient compute for processing all work and so a scale-up should be strongly considered. For simplicity this is only smoothed via a cooldown.
+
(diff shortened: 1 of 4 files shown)


docs/simulators/service-ewma/SPEC.md
```diff
@@ -0,0 +1,161 @@
+# Queue Consumer Scaling Algorithm Spec
+
+This document specifies the queue consumer scaling algorithm.
```
jaypipes: It seems odd to have the directory called `service-ewma` but the top heading of this file be `Queue Consumer Scaling Algorithm Spec`. Perhaps it's worth doing a few things here:

1. Change the name of the directory to `consumer-ewma` or `queue-poller-ewma`
2. Add a paragraph to the top of this file explaining that the algorithm described herein examines the number of consumers polling a task queue over a set time window and uses an Exponential Weighted Moving Average (EWMA) formula to recommend adjustments in the number of pollers.
3. Add a link or two to background documentation on EWMA.

docs/simulators/service-ewma/SPEC.md
```diff
@@ -0,0 +1,161 @@
...
+Each scaling decision emits one of:
+
+- No action.
+- Updated planned consumer count.
+
+The caller is responsible for reconciling running consumer processes to the updated planned consumer count and restoring that count on later decisions. If no prior planned consumer count is available, the caller uses `Initial consumer count` as the default.
+
+## Non-Goals
+
+The algorithm intentionally does not use:
```
jaypipes: I understand that Claude put a lot of this together, but it would be good to include an explanation of *why* the algorithm intentionally does not use these things in its evaluation. For instance, when I read the below list, I immediately asked myself "hold up... why wouldn't we want to take backlog length into account for the no-sync scale-up decision?"

jaypipes (changes requested):

docs/simulators/service-ewma/SPEC.md
02strich: Thanks, updated the folder name and the title of the doc (those were working titles). In a follow-up PR I'll update the remaining simulator folders to fit better.

Also added a bit of intro text to cover the high-level.

docs/simulators/service-ewma/SPEC.md
02strich: This is actually unrelated to Claude and more about this being a spec, which I originally started without any reasoning in it. That said, good feedback so added a bit more reasoning :)

jaypipes (approved):
Awesome, thanks @02strich :)

PR #55, opened 2026-05-28
Add worker set size metric and minor cleanups

02strich:
- Add SetWorkerSetSizeCount counter and increment it when handling UpdateWorkerSetSize actions.
- Use errors.AsType for NotFound check in DeleteWorkerControllerInstance and lowercase the task queue name error string.
- Drop duplicate enums import in hook.go and fix a stale doc comment in test_invoke.go.


jaypipes (approved):
👍 thanks @02strich :)

PR #58, opened 2026-06-03
Add deferred task-add scaling action

02strich:
Introduces a new `deferred-process-task-add` scaling action that lets algorithms request follow-up task-add processing with access to a metrics snapshot. The workflow handles the action by executing a new `HandleDeferredTaskAddAction` activity, which invokes the algorithm's `DeferredProcessTaskAdd` hook with a cached, task-type-filtered metrics snapshot getter.

Also adds:
- `DeferredProcessTaskAddCount` metric.
- Default no-op `DeferredProcessTaskAdd` implementation on the no-sync-match algorithm.
- Helpers in `scaling_metrics_snapshot.go` for filtering and caching metrics snapshots.
- Activity and algorithm tests covering the new path.

State updates from scaling responses are now applied before dispatching follow-up actions so deferred handlers see the latest status.

Background: for efficiency reasons the ProcessTaskAdd hook is run in a local activity. Unfortunately, that cannot access the metrics (as the metrics API is dependent on making a query to this same workflow so we get a circular dependency). The deferred hook allows offloading that, as well as any other higher latency processing into a “normal” activity.

### DIFF
--- a/wci/workflow/activities.go
+++ b/wci/workflow/activities.go
@@ -3,11 +3,12 @@ package workflow
 import (
 	"context"
 	"fmt"
+	"maps"
 	"slices"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
-	deploymentpb "go.temporal.io/api/deployment/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	workflowservice "go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/auto-scaled-workers/wci/client"
@@ -80,6 +81,22 @@ type (
 		Actions              []scalingalgorithm.ScalingAction        `json:"actions,omitempty"`
 	}
 
+	HandleDeferredScalingDecisionActivityRequest struct {
+		RequestContext
+
+		Request         iface.SignalTaskAddRequest `json:"request"`
+		ScalingGroupKey string                     `json:"scaling_group_key"`
+
+		ScalingGroupSpec   iface.ScalingGroupSpec       `json:"scaling_group_spec"`
+		EffectiveTaskTypes []enumspb.TaskQueueType      `json:"effective_task_types"`
+		ScalingStatus      iface.ScalingAlgorithmStatus `json:"scaling_status"`
+	}
+
+	HandleDeferredScalingDecisionActivityResponse struct {
+		UpdatedScalingStatus iface.ScalingAlgorithmStatus     `json:"scaling_status"`
+		Actions              []scalingalgorithm.ScalingAction `json:"actions,omitempty"`
+	}
+
 	PullStatsActivityRequest struct {
 		RequestContext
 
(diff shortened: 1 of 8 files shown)


wci/workflow/activities.go
```diff
@@ -468,7 +503,7 @@ func (a *Activities) PullStats(ctx context.Context, req *PullStatsActivityReques
 
 		logger.Debug("Loaded scaling algo", "scaling_algo", scalingAlgo, "config", scalingConfig)
 
-		response, err := scalingAlgo.ProcessMetricsPoll(ctx, scalingConfig, scalingStatus, scalingMetricsSnapshot)
+		response, err := scalingAlgo.ProcessMetricsPoll(ctx, scalingConfig, scalingStatus, *scalingMetricsSnapshot)
```
jaypipes, line 509: this is a nil-pointer exception waiting to happen since `filterScalingMetricsSnapshotByTaskTypes` can return nil. this needs a nil-guard check above after line 492.

wci/workflow/workflow.go
```diff
@@ -600,11 +615,63 @@ func (d *WorkflowRunner) handleActions(ctx workflow.Context, actions []scalingal
...
+				}).Counter(wcimetrics.Operations.Name()).Inc(1)
+			} else {
+				d.metrics.WithTags(map[string]string{
+					wcimetrics.OperationTagName: wcimetrics.OperationTypeDeferredTaskAdd,
+				}).Counter(wcimetrics.Operations.Name()).Inc(1)
+
+				if resp.UpdatedScalingStatus != nil {
+					d.State.ScalingStatus[action.ScalingGroupKey] = resp.UpdatedScalingStatus
+				}
+				d.handleActions(ctx, resp.Actions, nil)
```
jaypipes, line 668: So what happens if the controller fails between line 667 and 669?

wci/workflow/workflow.go
```diff
@@ -580,14 +583,26 @@ func (d *WorkflowRunner) handleNoSyncMatchSignal(ctx workflow.Context, req *ifac
...
 
-func (d *WorkflowRunner) handleActions(ctx workflow.Context, actions []scalingalgorithm.ScalingAction) {
+// handleActions dispatches scaling actions returned by the scaling algorithm.
+//
+// ScalingStatus is treated as intent, not confirmation: callers must persist
+// resp.UpdatedScalingStatus into d.State.ScalingStatus before invoking this
+// function. The deferred-task-add case forwards d.State.ScalingStatus to its
+// follow-up activity and so must see the freshly-computed status. If a
+// dispatched action subsequently fails, the persisted status is not rolled
+// back — the algorithm sees the optimistic state on the next iteration and
```
jaypipes: It's not really "reconciling drift" -- at least in the way that, say, a Kubernetes controller reconciles desired state -> observed state drift. What the handleActions() is doing, rather, is simply updating the latest workflow status/state with whatever value was in the `HandleDeferredTaskAddAction` activity execution response's UpdatedScalingStatus...

jaypipes (changes requested):
I'd prefer that the activity be named something a lot more specific. Instead of `HandleDeferredTaskAddAction`, can we name it `PullScalingMetricsSnapshot` so that it's more clear what the activity is actually doing?

wci/workflow/activities.go
02strich: good catch, added

wci/workflow/workflow.go
02strich: yeah, I had redone this description shorter and longer a few times and was never quite happy... I have removed the last part as it is not that (as you say) and seems to confuse/mislead without providing value.

What this describes is the need for it to see the freshest ScalingStatus available to be consistent - and so that the algorithm output already needs to have been stored into the workflow state.

wci/workflow/workflow.go
02strich: lol, thanks - moving the metric to the right place

jaypipes (approved):
thanks for making those updates :)

PR #59, opened 2026-06-17
Adding algorithm for GCP CloudRun WorkerPools and similar compute providers

02strich:
## What was changed
Introduce a rate-based scaling algorithm that sizes the worker pool from EWMA-smoothed arrival, dispatch, and per-consumer capacity estimates. Desired worker count is the maximum of a utilization-target model and a Halfin-Whitt
square-root staffing model, bounded by min/max count, capped per scale-up step, and gated by asymmetric scale-up/scale-down cooldowns plus a no-sync-match quiet period. Includes unit tests covering the sizing and scaling decisions

This is an implementation of the algorithm in https://github.com/temporalio/temporal-auto-scaled-workers/pull/54

## Why?
The existing scaling algorithm makes a set of assumption that are true for Lambda, but not for other compute providers like GCP CloudRun or AWS ECS. This adds a new scaling algorithm more suited to these other compute providers.

### DIFF
--- a/wci/workflow/iface/map_access.go
+++ b/wci/workflow/iface/map_access.go
@@ -1,6 +1,7 @@
 package iface
 
 import (
+	"math"
 	"strconv"
 
 	"go.temporal.io/api/serviceerror"
@@ -114,6 +115,9 @@ func validateFloat64InMap(m map[string]any, key string, minValidValue float64) e
 	case float64:
 		if val < minValidValue {
 			return serviceerror.NewInvalidArgumentf("%s must be at least %v", key, minValidValue)
+		}
+		if math.IsInf(val, 0) || math.IsNaN(val) {
+			return serviceerror.NewInvalidArgumentf("%s must be a real float", key)
 		}
 	case string:
 		i, err := strconv.ParseFloat(val, 64)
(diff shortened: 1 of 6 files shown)


wci/workflow/iface/map_access.go
```diff
@@ -115,6 +116,9 @@ func validateFloat64InMap(m map[string]any, key string, minValidValue float64) e
 		if val < minValidValue {
 			return serviceerror.NewInvalidArgumentf("%s must be at least %v", key, minValidValue)
 		}
+		if math.IsInf(val, 0) || math.IsNaN(val) {
+			return serviceerror.NewInvalidArgumentf("%s must be a real float", key)
+		}
```
gcristea-temporal, line 121: Nit: I think this check should happen before the check against `minValue`. Maybe first convert to `float64` and checking would make this easier to read.

gcristea-temporal (approved):
Looks good, only found one nit.

gcristea-temporal:
$\int or\mu las$

wci/workflow/iface/map_access.go
02strich: Out of curiosity, why do you think it matters? (b/c of the -inf case?)

gcristea-temporal: I was thinking about the case where `val` is `NaN`. For that situation the first check (line 116) will return the error.

PR #60, opened 2026-06-19
Add configurable validation interval and connectivity health status

smuneebahmad:
## What was changed
 WCI now signals the version workflow whenever its connectivity validation status changes (on config save, periodic check, "Validate Connection", or a scaling invocation failure). This enables the deployment workflow to maintain an up-to-date aggregate connectivity summary in its memo

## Why?
Enables surfacing the connectivity health between Temporal and a customer's compute resource (e.g. Lambda) through the Worker Deployments API, so the UI can show ongoing validations status of compute configs without additional queries.

## Checklist

1. Closes   https://temporalio.atlassian.net/browse/COM-143

2. How was this tested:
Deployed to a test cell and verified that connectivity status updates correctly after triggering "Validate Connection" and after a failed Lambda invocation. Confirmed the version workflow receives the signal and the deployment workflow memo is updated accordingly.


wci/workflow/compute_provider/aws_lambda.go
```diff
@@ -58,17 +58,17 @@ func (p *awsLambdaComputeProvider) ValidateConfig(ctx context.Context, cfg Compu
...
 	if _, err = lambdaClient.GetFunction(ctx, &lambda.GetFunctionInput{
 		FunctionName: aws.String(arn),
 	}); err != nil {
-		return fmt.Errorf("lambda GetFunction failed: %w", err)
+		return fmt.Errorf("Can't access the compute resource: %w", err)
 	}
 
 	if err := p.checkExternalID(ctx, cfg, arn); err != nil {
-		return err
+		return fmt.Errorf("Trust credential not enforced on the role: %w", err)
```
02strich: What's a trust credential? I think you can be a bit more explicit here

wci/workflow/workflow.go
```diff
@@ -290,11 +303,15 @@ func (d *WorkflowRunner) handleValidateSpec(ctx workflow.Context, args *iface.Va
 			}).Counter(wcimetrics.Updates.Name()).Inc(1)
 
 			if appErr, ok := errors.AsType[*temporal.ApplicationError](err); ok {
+				d.State.ValidationStatus = iface.NewValidationStatusFailed(validationTime, appErr.Message())
+				d.signalVersionWorkflow(ctx)
```
02strich: Why? The change did not get applied

wci/workflow/workflow.go
```diff
@@ -290,11 +303,15 @@ func (d *WorkflowRunner) handleValidateSpec(ctx workflow.Context, args *iface.Va
...
 				return nil, serviceerror.NewInvalidArgumentf("%s", appErr.Message())
 			} else {
 				return nil, err
 			}
 		}
+		d.State.ValidationStatus = iface.NewValidationStatusSuccess(validationTime)
+		d.signalVersionWorkflow(ctx)
```
02strich: No change occurred...

wci/workflow/iface/workflow.go
```diff
@@ -38,6 +38,10 @@ const (
 	// Signals
 	SignalTaskAdd = "task-add-signal"
 
+	// Signal sent to the version workflow when ValidationStatus changes, so the
+	// deployment workflow can maintain an aggregate validation summary in its memo.
+	SignalSyncValidationStatus = "sync-validation-status"
```
02strich: Seems odd to define this here, given it's not wcis interface

wci/workflow/workflow.go
```diff
@@ -194,9 +201,14 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 	addStatsPullTimer(maxPollInterval)
 
 	if d.hasMinVersion(PeriodicValidationVersion) {
+		validationInterval := getPeriodicValidationInterval(ctx, d.unsafePeriodicValidationIntervalGetter)
```
02strich: Are you OK with this setting only changing on can boundaries?

02strich (changes requested):

wci/workflow/workflow.go
smuneebahmad: The spec being validated here is the currently applied config. `handleValidateSpec()` intentionally updates ValidationStatus because the connectivity check result is a meaningful health signal of the current config's connectivity. In `L307` here, it indicates that connectivity validation for the current applied config failed, and `L314` below indicates the connectivity validation for the current applied config succeeded.

The new config change only gets applied as part of the `handleUpdateInstance()` fn.

wci/workflow/workflow.go
smuneebahmad: Same explanation as above on `L307`

wci/workflow/workflow.go
smuneebahmad: Yes, I think for the short-term this is reasonable, with the user having an option available to manually trigger the validation on-demand from the UI. As a follow up, we can provide an additional signal option to trigger a CaN on-demand.

wci/workflow/iface/workflow.go
smuneebahmad: Agreed it's not part of WCI's own interface; moved to `wci/client/`

wci/workflow/compute_provider/aws_lambda.go
smuneebahmad: Updated to be more prescriptive

wci/client/config.go
```diff
@@ -65,4 +69,9 @@ var (
 		60_000, // 1 minute
 		`WorkerControllerMinSignalIntervalSyncMatchMilliseconds controls the batching interval of sync matches grouped by WCI in milliseconds (per namespace). Each batch triggers a signal.`,
 	)
+	WorkerControllerPeriodicValidationIntervalMilliseconds = dynamicconfig.NewGlobalIntSetting(
+		"workercontroller.periodic_validation_interval_ms",
```
jaypipes: Hmm... should we make this setting seconds instead of milliseconds? Is a sub-second validation interval ever something we would recommend?

wci/workercomponent/component_test.go
```diff
@@ -0,0 +1,386 @@
+package workercomponent_test
```
jaypipes: This is a good additional set of unit tests. However, unless I'm mistaken, none of the unit tests are validating code that was introduced in *this particular PR*. So, this really should be added as a separate independent PR...

wci/workflow/workflow.go
```diff
@@ -194,9 +202,17 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 	addStatsPullTimer(maxPollInterval)
 
 	if d.hasMinVersion(PeriodicValidationVersion) {
+		// Read once at run start. Dynamic config changes take effect at the next
+		// ContinueAsNew. The interval defaults to 6h and is
+		// unlikely to change frequently, so waiting for a CaN boundary is acceptable.q
```
jaypipes:
```suggestion
		// unlikely to change frequently, so waiting for a CaN boundary is acceptable.
```

wci/workflow/workflow.go
```diff
@@ -79,7 +86,7 @@ type (
 // history clean so that we have less concern about backwards and forwards compatibility.
 // In steady state (i.e. absence of ongoing updates or signals) the wf should only have
 // a single wft in the history.
-func Workflow(ctx workflow.Context, unsafeWorkflowVersionGetter func() WorkerControllerInstanceWorkflowVersion, unsafeMaxVersion func() int, args *iface.WorkerControllerInstanceWorkflowArgs, activities *Activities) error {
+func Workflow(ctx workflow.Context, unsafeWorkflowVersionGetter func() WorkerControllerInstanceWorkflowVersion, unsafeMaxVersion func() int, unsafePeriodicValidationIntervalGetter func() time.Duration, args *iface.WorkerControllerInstanceWorkflowArgs, activities *Activities) error {
```
jaypipes: If we want to have consistent naming (recommended), the newly-added parameter should be `unsafePeriodicValidationInterval`, not `unsafePeriodicValidationIntervalGetter`, since `unsafeMaxVersion` isn't suffixed with `Getter` (even though it's a getter :) ).

jaypipes (commented):
Overall good stuff. Just some minor comments inline...

wci/workercomponent/component_test.go
smuneebahmad: Yes, not directly related to the current PR. Will separate out to a new PR.

wci/workflow/workflow.go
smuneebahmad: Makes sense. Updated in the next patch.

wci/workflow/workflow.go
smuneebahmad: Thanks for catching this :)

wci/client/config.go
smuneebahmad: Changed to seconds

jaypipes (approved):
👍 thanks @smuneebahmad !

wci/workflow/iface/workflow.go
02strich: I think this should be defined on the temporalio/temporal repo which this already depends on

wci/workflow/workflow.go
02strich: > The spec being validated here is the currently applied config

not as far as I am reading the code. The spec validated is the current spec with the provided changes applied in L270.

02strich (changes requested):

02strich (approved):

PR #61, opened 2026-06-23
COM-163: Added support for GCP impersonation chain.

gcristea-temporal:
<!--- Note to EXTERNAL Contributors -->
<!-- Thanks for opening a PR! 
If it is a significant code change, please **make sure there is an open issue** for this. 
We work best with you when we have accepted the idea first before you code. -->

<!--- For ALL Contributors 👇 -->

## What was changed
The GCP Cloud Run compute provider needs the customer namespace to resolve the per-namespace IAM impersonation chain. This PR threads that identity through the `ComputeProvider` interface and makes the chain resolution pluggable.

### Changes                                                                                                                                                           
                                                                                                                                                                       
  - **`ComputeProvider` interface** now takes an `InvocationContext{Namespace, DeploymentName, BuildID}` instead of a bare config. The struct is owned by the`compute_provider` package and extends without further signature churn. All providers (AWS ECS/Lambda, GCP Cloud Run, k8s, subprocess, test fakes) updated.                                                                                                               
  - **Pluggable impersonation chain.** GCP Cloud Run resolves its impersonation delegates through a process-wide `ImpersonationChainProvider`, defaulting to `NoopImpersonationChainProvider` (which reproduces the prior behavior: random SA pick per hop). The real per-namespace resolver can be installed via `fx.Decorate` + `SetImpersonationChainProvider`, wired in `wci/fx.go`.
  - **Activity layer** builds the `InvocationContext` from each request's embedded `RequestContext` (`req.invocationContext()`).
  - **Validation:** `ValidateSpecWorkflow` (standalone dry-run path) was constructing its `ValidateSpecRequest` without a `RequestContext`, so `ValidateConfig` saw an empty namespace. It now populates it from `workflow.GetInfo(ctx).Namespace` (the workflow runs in the customer namespace). Without this, validation would resolve a different impersonation chain than the runtime `InvokeWorker`/`UpdateWorkerSetSize` path once a real resolver is installed.                                                                                       
                                                                                                                                                                       
  ### Compatibility / risk                                                                                                                                              
                                                                                                                                                                       
  - Default behavior is unchanged: the Noop provider preserves the existing random-per-hop delegate selection.                                                                                                                                 
  - The chain provider is a mutable package singleton set once at startup. Known tradeoff (tests mutate it with `t.Cleanup`, so they can't run `t.Parallel`);
                                                                                                                                                                       
## Why?
<!-- Tell your future self why have you made these changes -->
Support for GCP Cloud Run pre-release

## Checklist
<!--- add/delete as needed --->

1. Closes <!-- add issue number here -->

2. How was this tested:
<!--- Please describe how you tested your changes/how we can test them -->

3. Any docs updates needed?
<!--- update README if applicable
      or point out where to update docs.temporal.io -->

### DIFF
--- a/wci/workflow/activities.go
+++ b/wci/workflow/activities.go
@@ -40,13 +40,11 @@ type (
 		workflowserviceClient workflowservice.WorkflowServiceClient
 	}
 
-	// RequestContext carries the namespace/deployment identity tags shared by every
-	// activity request type.
-	RequestContext struct {
-		NamespaceName     string `json:"namespace_name"`
-		DeploymentName    string `json:"deployment_name"`
-		DeploymentBuildID string `json:"deployment_build_id"`
-	}
+	// RequestContext aliases the compute-provider request context so activity
+	// request types can embed it while compute providers consume the same type
+	// directly. The alias keeps it defined in compute_provider (which the
+	// ComputeProvider interface depends on) without an import cycle.
+	RequestContext = computeprovider.RequestContext
 
 	ValidateSpecRequest struct {
 		RequestContext
@@ -116,23 +114,15 @@ type (
 	}
 )
 
-// invocationContext projects the request's identity into the form the compute
-// providers consume.
-func (r RequestContext) invocationContext() computeprovider.InvocationContext {
-	return computeprovider.InvocationContext{
-		Namespace:      r.NamespaceName,
-		DeploymentName: r.DeploymentName,
-		BuildID:        r.DeploymentBuildID,
-	}
-}
-
 // metricsHandler returns the activity-side MetricsHandler pre-tagged with the
-// namespace/deployment identity and the supplied activity type.
-func (r RequestContext) metricsHandler(ctx context.Context, activityType wcimetrics.ActivityType) sdkclient.MetricsHandler {
+// namespace/deployment identity and the supplied activity type. It is a free
+// function because RequestContext is an alias of a compute_provider type, so
+// methods can't be declared on it from this package.
+func metricsHandler(ctx context.Context, rc RequestContext, activityType wcimetrics.ActivityType) sdkclient.MetricsHandler {
 	return activity.GetMetricsHandler(ctx).WithTags(map[string]string{
-		wcimetrics.NamespaceTag:               r.NamespaceName,
-		wcimetrics.WorkerDeploymentNameTag:    r.DeploymentName,
-		wcimetrics.WorkerDeploymentBuildIDTag: r.DeploymentBuildID,
+		wcimetrics.NamespaceTag:               rc.NamespaceName,
+		wcimetrics.WorkerDeploymentNameTag:    rc.DeploymentName,
+		wcimetrics.WorkerDeploymentBuildIDTag: rc.DeploymentBuildID,
 		wcimetrics.ActivityTypeTag:            string(activityType),
 	})
 }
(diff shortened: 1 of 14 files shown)


wci/workflow/activities.go
```diff
@@ -116,6 +116,16 @@ type (
...
 
+// invocationContext projects the request's identity into the form the compute
+// providers consume.
+func (r RequestContext) invocationContext() computeprovider.InvocationContext {
+	return computeprovider.InvocationContext{
+		Namespace:      r.NamespaceName,
+		DeploymentName: r.DeploymentName,
+		BuildID:        r.DeploymentBuildID,
+	}
+}
```
jaypipes: Is there a reason to create a separate InvocationContext struct as opposed to simply using the existing RequestContext struct?

wci/workflow/compute_provider/aws_lambda.go
```diff
@@ -46,7 +46,7 @@ func (p *awsLambdaComputeProvider) LaunchStrategy() LaunchStrategy {
 	return LaunchStrategyInvoke
 }
 
-func (p *awsLambdaComputeProvider) ValidateConfig(ctx context.Context, cfg ComputeProviderConfig) error {
+func (p *awsLambdaComputeProvider) ValidateConfig(ctx context.Context, _ InvocationContext, cfg ComputeProviderConfig) error {
```
jaypipes: The comment in the PR summary states:

> Validation: ValidateSpecWorkflow (standalone dry-run path) was constructing its ValidateSpecRequest without a RequestContext, so ValidateConfig saw an empty namespace. It now populates it from workflow.GetInfo(ctx).Namespace (the workflow runs in the customer namespace). Without this, validation would resolve a different impersonation chain than the runtime InvokeWorker/UpdateWorkerSetSize path once a real resolver is installed.

Does the fact that we are ignoring the `InvocationContext` parameter here in the lambda compute provider imply that the customer namespace is not relevant for the Lambda compute provider validation? Or, does it imply that the customer namespace is not relevant for *any* compute provider that uses the Invoke strategy? Or does it imply neither? :)

jaypipes (changes requested):
I don't think there's a need for a new InvocationContext struct. I'd prefer we simply pass the existing RequestContext struct which has all the information that InvocationContext has... Also, see another question inline. Thanks @gcristea-temporal!

wci/workflow/compute_provider/gcp_cloudrun.go
```diff
@@ -131,3 +126,73 @@ func (p *gcpCloudRunComputeProvider) buildClientAndParams(ctx context.Context, c
...
+	}
+	workerPool, ok := config[configGCPCloudRunWorkerPool].(string)
+	if !ok || workerPool == "" {
+		return "", fmt.Errorf("worker_pool not found in config")
+	}
+	return fmt.Sprintf("projects/%s/locations/%s/workerPools/%s", project, region, workerPool), nil
+}
+
+type (
+	ImpersonationChainProvider interface {
```
02strich: given that this is GCP specific, should the name reflect that? I also would probably split out of this file as it is related to GCP not just CloudRun (and this grows the amount of code)

02strich:
+1 to Jay's feedback.

My original idea was to do the pre-processing outside the compute provider. That said, the approach from this PR might be nicer as more type-safe and I do have another use case where I believe to need the namespace in the compute provider...

wci/workflow/activities.go
gcristea-temporal: I introduced `InvocationContext` mainly because using `RequestContext` in the `compute_provider` can't be done with refactoring: it is currently defined in `workflow` which depends on `compute_provider`. 

Having code from `compute_provider` use `RequestContext` requires it moving to `compute_provider` (or to a new package.) 

The problem with moving it (the way it was defined) is with the [metricsHandler](https://github.com/temporalio/temporal-auto-scaled-workers/blob/main/wci/workflow/activities.go#L121) which would bring in a dependency on `wci/metrics` into `compute_provider`. 

Refactored to use `RequestContext`:
* Moved its definition to `compute_provider`
* Using a type alias in `workflow`
* Changed `metricsHandler` as a free function (instead of a method on `RequestContext`)

@jaypipes @02strich

wci/workflow/compute_provider/gcp_cloudrun.go
gcristea-temporal: Renamed to `GCPImpersonationChainProvider`.

jaypipes (approved):
👍 thanks @gcristea-temporal !

PR #62, opened 2026-06-24
Automated Testing

tylercunnion:
## What was changed
Adds automated testing on main and PR branches.

### Linting
Adds a golangci-lint configuration synthesized from other core repos in the organization. Linters run only on code changed within the instant pull request. Linters do not run on the main branch.

### Unit Testing
Unit tests run on all PR and main builds.

We collect but do not currently do anything with code coverage data. In the future we may want to integrate with Codecov, in line with some of our other projects.

### Integration Testing
Integration tests use the [server test environment from the OSS server repo](https://github.com/temporalio/temporal/tree/main/tests/testcore), with the WCI component injected in. Integration tests are run against both the pinned version of the server (as defined in `go.mod`) as well as the current `main` branch (these are advisory and will not fail the build). 

Currently the only defined integration test is a simple smoke test to ensure we can create a worker deployment version, this will be expanded in future work.

Integration tests live in a nested submodule for dependency separation etc; there is a CI job configured to enforce that the versions defined in the respective `go.mod` files for the go language version as well as the temporal server versions stay in sync. This check will fail the build, in order to prevent drift.

## Why?

Enforces code standards as well as ensuring ongoing validation for the codebase.


Makefile
```diff
@@ -36,7 +37,7 @@ clean-test-output:
 
 unit-test: clean-test-output
 	@printf $(COLOR) "Run unit tests..."
-	@CGO_ENABLED=$(CGO_ENABLED) go test $(UNIT_TEST_DIRS) $(COMPILED_TEST_ARGS) 2>&1 | tee -a test.log
+	@CGO_ENABLED=$(CGO_ENABLED) go test $(UNIT_TEST_FLAGS) $(UNIT_TEST_DIRS) $(COMPILED_TEST_ARGS) 2>&1 | tee -a test.log
```
02strich, line 40: what sets `UNIT_TEST_FLAGS`?

tests/go.mod
02strich, line 1: why do tests get their own `go.mod`?

tests/.gitignore
02strich, line 1: ??

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,114 @@
...
+            "$(grep 'go\.temporal\.io/server ' go.mod | awk '{print $2}')" \
+            "$(grep 'go\.temporal\.io/server ' tests/go.mod | awk '{print $2}')"
+
+          exit $fail
+
+  unit-test:
+    permissions:
+      contents: read
+      checks: write
+      statuses: write
```
02strich: what is this needed for?

tests/integration/wci_test.go
```diff
@@ -0,0 +1,57 @@
...
+		nil,
+	)
+	require.NoError(t, err)
+
+	exists, err = env.Client.WorkerControllerInstanceExists(ctx, nsEntry, version)
+	require.NoError(t, err)
+	require.True(t, exists)
+
+	err = env.Client.DeleteWorkerControllerInstance(ctx, nsEntry, version, "test-identity")
+	require.NoError(t, err)
```
02strich: should there be another exists check after this?

Makefile
tylercunnion: Nothing in the automated tests (right now), this is for someone invoking `make` manually and wants to add a flag to `go test`.

.github/workflows/ci.yaml
tylercunnion: At the moment nothing, they would be needed if we wanted to publish code coverage etc. I can remove them until then.

tests/.gitignore
tylercunnion: Yeah that's an artifact from something I was trying to do, not needed

Makefile
02strich: so just for completeness? cool

tests/integration/wci_test.go
tylercunnion: Works for me yeah

tests/go.mod
tylercunnion: Basically trying to isolate them so that a) the dependencies that are there just for integration tests don't leak into the main module; b) it makes the CI a little more straightforward when we want to run integration against two different versions of temporal server.

That said, since I went back on the original idea of actually mocking out portions of the AWS API, I don't think I'm going to end up introducing anything huge into the dependency graph like a Testcontainers or anything. Let me play around with what it would look like to simplify it

jaypipes: It's generally good practice for testing-only packages to have their own `go.mod`, so I support this approach. That said, I'd make the package `internal/testing` instead of `tests` at the top-level.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,112 @@
...
+        run: go test -tags=test_dep -timeout 300s ./integration/...
+
+  integration-test-server-main:
+    continue-on-error: true
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          submodules: recursive
```
jaypipes: The submodules here refers to *Git* submodules, not Go modules. Unless you have a very good reason, I'd avoid Git submodules like the plague...

jaypipes (commented):
👍 The integration tests are a lot more valuable than the unit tests (IMHO) and generally easier to reason about and extend. So, +1 from me.

.github/workflows/ci.yaml
tylercunnion: Ah yeah good call. No reason for git submodules in here at all, I'll nuke that

jaypipes (approved):

tests/go.mod
tylercunnion: I think for now I'm going to leave them where they're at - `internal` makes more sense if we didn't do the multi-module approach, since that would prevent what's in there from being imported elsewhere; as it is though, since the tests are in their own module, they're already not importable and we'd just be nesting for not a ton of benefit IMO

02strich: I don't quite see how this is a 'testing-only package' as it is more the testing package for the module it is in. @jaypipes looking to learn more: what is the motivation for the practice/what is the problem it is trying to solve?

wci/workflow/compute_provider/test_invoke.go
```diff
@@ -18,33 +19,79 @@ const (
...
+var (
+	invokeObserverMu sync.RWMutex
+	// invokeObservers maps a deployment build ID to the observer watching it,
+	// so concurrent tests each observe only their own build's actions.
+	invokeObservers = map[string]InvokeObserver{}
+)
+
+// emitProviderEvent reports an action to the observer registered for the
+// request's deployment build, if any.
+func emitProviderEvent(rc RequestContext, action string) {
```
02strich, line 86: should this have a more specific name to avoid it colliding with non-test code?

wci/workflow/compute_provider/test_invoke_observer.go
02strich, line 1: why is this in it's own file?

tylercunnion: So that we can put the `test_dep` build tag on and prevent non-test code from attaching an observer here. My preference would probably have been to just have a separate ComputeProvider implementation altogether but it has to pass the validity check (be part of the enum of available Providers) so we can't just put an arbitrary one in there.

wci/workflow/compute_provider/test_invoke.go
tylercunnion: Yeah I think we can make this `emitTestProviderEvent` which should avoid any potential collision and be a little more specific about intent

wci/workflow/compute_provider/test_invoke_observer.go
02strich: hrm, I am not sold on the value of that split today and do agree if we get really concerned we could invest more into having the test providers really only be in the test runs. That said, they are not exposed in TCloud and people can remove them if they like in their OSS setups

PR #63, opened 2026-06-24
Increase PullStats timeout

02strich:
## What was changed
Moving from 15 to 60 seconds

## Why?
We are seeing a lot of cases of timeouts, and it is not that critical

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/wci/workflow/workflow.go
+++ b/wci/workflow/workflow.go
@@ -19,7 +19,7 @@ import (
 
 const (
 	ValidateSpecActivityTimeout                  = 15 * time.Second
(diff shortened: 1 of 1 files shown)


ericbriffa:
> and it is not that critical

can you share what the timeouts are and why they are not critical?

02strich:
@ericbriffa it was 15s and this moves it to 60s. The pullstats is a background path, not in the interactive experience of the customer

smuneebahmad (approved):

smuneebahmad (approved):

PR #65, opened 2026-06-29
Move persistence store providers to worker DI

02strich:
## What was changed
Move DataStoreFactoryProvider and the metadata/cluster-metadata store providers out of the wci Module and into the worker's fx graph, so newMetadataManager and newClusterMetadataManager now receive the stores directly instead of constructing them from the DataStoreFactory.

## Why?
To allow more configurability around the data store implementation, the WCI module now only depends on the specific data store parts it actually uses instead of the whole factory interface.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -5,8 +5,8 @@ import (
 	"os"
 	_ "time/tzdata" // embed tzdata as a fallback
 
-	"go.temporal.io/auto-scaled-workers/wci"
 	"github.com/urfave/cli/v2"
+	"go.temporal.io/auto-scaled-workers/wci"
 	"go.temporal.io/server/common/build"
 	"go.temporal.io/server/common/config"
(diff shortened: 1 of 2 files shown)


jaypipes (approved):

cmd/worker/main.go
```diff
@@ -118,6 +119,8 @@ func buildCLI() *cli.App {
 					config.Module,
 					temporal.FxLogAdapter,
 					ringpop.MembershipModule,
+					fx.Provide(persistenceClient.DataStoreFactoryProvider),
```
smuneebahmad, line 122: Should we also add `fx.Invoke(persistenceClient.DataStoreFactoryLifetimeHooks)` here?

02strich: you are right, good catch

smuneebahmad (approved):

PR #66, opened 2026-07-01
Add Lambda unit tests

tylercunnion:
## What was changed
Add unit tests for Lamba Invoke and ValidateConfig logic.
Since the logic for these requires a Lambda Client, this change adds a package-level variable (the same as we currently have for `verifyExternalIDEnforced`) that allows us to swap in, on a test-only basis, a function to provide our own struct that implements the minimal interface required here.

## Why?
Verifying that the AWS Lambda API is called in the way we expect when receiving an Invoke signal is crucial to the overall test strategy. Since integration tests will check only that a TQ backlog leads to a (non-provider-specific) Invoke call, we need the unit tests to verify that those Invoke calls do what we expect.


jaypipes (approved):
These are fine since they follow the existing unit test pattern. I'm not sure the stubbed/mocked unit tests provide a whole lot of value compared to integration tests, but OK :)

jaypipes (approved):

tylercunnion:
> These are fine since they follow the existing unit test pattern. I'm not sure the stubbed/mocked unit tests provide a whole lot of value compared to integration tests, but OK :)

Yeah the main value they provide is the "last mile" validation - integration test proves out every part of the WCI machinery up to calling "Invoke" etc on the compute provider. Since we don't have a real compute provider wired up in the integration test, this will prove that each individual provider responds correctly to the Invoke call. So it's a smaller portion of the overall picture but still important in their own right

PR #68, opened 2026-07-01
Add Linting and Formatting targets in Make, fail CI on lint/format

tylercunnion:
## What was changed

Changes the "Linting" github action to enforce the linter output. That includes formatting.

To compensate, this also adds new make targets:

* `make lint` - lints the entire project
*  `make lint-branch` - lints just the current branch, using the merge base with `main`
* `make fmt` - reformats the entire project according to the rules in `golangci-lint.yml`
* `make fmt-branch` - reformats but only changed files on this branch. useful for cleaning up to pass the CI check.

## Why?
The CI check is initially advisory, but it's so easy to ignore that it's not worth having. This makes it mandatory while also making it easy to get the format right.


wci/just_to_check.go
02strich, line 1: once this is removed, I am ready to approve the PR :)

02strich (approved):

PR #70, opened 2026-07-02
Record validation status when validating current spec

02strich:
## What was changed
In handleValidateSpec, update State.ValidationStatus and signal the version workflow on both success and failure, but only when no scaling group changes are submitted (i.e. validating the current configuration rather than a dry-run of changes).

## Why?
There was confusion on the meaning of the validate spec API, as people expected it to support validating the current config as well - which was not the original thought. This changes the implementation with expectations.


jaypipes (approved):
Ausgezeichnet, vielen Dank!

PR #73, opened 2026-07-06
Update GCP mask to align with docs

02strich:
Updating the update mask for GCP cloudrun to align with https://docs.cloud.google.com/run/docs/configuring/workerpools/manual-scaling#rest-api

## What was changed
Update the update mask for GCP CloudRun to be specific to the instance count, not scaling overvall.

## Why?
To explicitly align with the GCP docs as that was provided as guidance.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


jaypipes (approved):
👍

PR #74, opened 2026-07-06
Add Nexus-based compute providers for worker invocation and scaling

02strich:
Introduce a Nexus RPC compute providers for invoke and worker-set compute providers.

As the Nexus-based compute providers are called directly from the workflow, adding a new code path to the Workflow to do that - and fallback to activity when another compute provider is used. Also backfilling the so far missing field for the Nexus endpoint - which already existed in the frontend API.

Add a cmd/nexus-subprocess-example binary demonstrating a subprocess-backed Nexus invoke service, along with Makefile targets and .gitignore entries for the generated files and new binary.


cmd/nexus-subprocess-example/main.go
bergundy, line 1: I would call out in a comment that this is temporary until we add generate language friendly types with nexgen.

wci/workflow/compute_provider/nexus/internal/nexusservergen/main.go
```diff
@@ -0,0 +1,237 @@
...
+	flag.Parse()
+
+	if *packageName == "" {
+		fatalf("--package is required")
+	}
+	if *outFile == "" {
+		fatalf("--out-file is required")
+	}
+	if flag.NArg() != 1 {
+		fatalf("expected one .nexusrpc.yaml input file")
```
bergundy, line 60: You shouldn't need a nexusrpc.yaml for proto based definitions.

bergundy: This should all be handled by nexgen.

wci/workflow/compute_provider/nexus/temporal/api/common/v1/message.proto
bergundy, line 1: You should be able to avoid copying this definition into your repo. I think you should define all of your protos in the api repo.

wci/workflow/compute_provider/nexus/generate.go
```diff
@@ -0,0 +1,12 @@
+package nexus
+
+// temporal/api/common/v1/message.proto is a vendored, hand-maintained subset of the
+// upstream go.temporal.io/api common message definitions, checked in only so that
+// `protoc --proto_path=.` can resolve the Payload import. It must stay a faithful
+// subset of the upstream definition; update it if the upstream Payload message changes.
+
+//go:generate protoc --proto_path=. --go_out=. --go_opt=paths=source_relative nexus.proto nexus_invoke.proto nexus_worker_set.proto
+//go:generate npx --yes nexus-rpc-gen@0.1.0-alpha.4 --lang go --package nexus --out-file nexus_invoke_nexusrpc_gen.go nexus_invoke.nexusrpc.yaml
+//go:generate npx --yes nexus-rpc-gen@0.1.0-alpha.4 --lang go --package nexus --out-file nexus_worker_set_nexusrpc_gen.go nexus_worker_set.nexusrpc.yaml
```
bergundy, line 10: nexus-rpc-gen is deprecated. We should sync up on this.

wci/workflow/compute_provider/nexus/nexus_worker_set.go
```diff
@@ -0,0 +1,104 @@
...
+	return nil
+}
+
+func (p *nexusWorkerSetComputeProvider) newNexusClient() workflow.NexusClient {
+	return newNexusWorkerSetWorkflowClientFn(p.endpoint, WorkerSetComputeProvider.ServiceName)
+}
+
+var newNexusWorkerSetWorkflowClientFn = workflow.NewNexusClient
+
+var executeNexusWorkerSetValidateConfigOperationFn = func(
```
bergundy, line 82: There's a plan to generate client methods in nexgen too, which will save you from creating these wrappers.

wci/workflow/compute_provider/nexus/temporal/api/common/v1/message.proto
02strich: I was wondering about that - given it's a user facing API that seems fitting. I'll see to move them over there

PR #75, opened 2026-07-06
Add NexusEndpoint field to ComputeProviderSpec

02strich:
## What was changed
Adding the missing nexus endpoint field to the config struct

## Why?
In preparation for adding support for Nexus-based providers, adding the endpoint field on the WCI end. This already exists in the API so this is just backfill.


jaypipes (approved):
I don't know what a "Nexus-based provider" is, but ok.

PR #76, opened 2026-07-06
COM-186: Migrate dynamic config from cell/global to namespace.

gcristea-temporal:
## What was changed
Moved the `workercontroller.compute_providers.enabled` config value from the global/cell priority to namespace priority. 

## Why?

This change allows to specify a global value and then override it at the namespace level.
## Checklist

1. Closes 
[COM-186](https://temporalio.atlassian.net/browse/COM-186)
2. How was this tested:

* Added unit tests for `wci/client/config`
3. Any docs updates needed?

* No

[COM-186]: https://temporalio.atlassian.net/browse/COM-186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wci/client/config.go
+++ b/wci/client/config.go
@@ -18,6 +18,7 @@ type (
 	}
 )
 
+// Namespace-scoped settings.
 var (
 	WorkerControllerEnabled = dynamicconfig.NewNamespaceBoolSetting(
 		"workercontroller.enabled",
@@ -34,6 +35,25 @@ var (
 		2,
 		`WorkerControllerInstanceWorkflowVersion controls what version of the logic should the manager workflows use.`,
 	)
+	WorkerControllerMinSignalIntervalNoSyncMatchMilliseconds = dynamicconfig.NewNamespaceIntSetting(
+		"workercontroller.hook.min_signal_interval_no_sync_match",
+		500,
+		`WorkerControllerMinSignalIntervalNoSyncMatchMilliseconds controls the batching interval of no-sync matches grouped by WCI in milliseconds (per namespace). Each batch triggers a signal. `,
+	)
+	WorkerControllerMinSignalIntervalSyncMatchMilliseconds = dynamicconfig.NewNamespaceIntSetting(
+		"workercontroller.hook.min_signal_interval_sync_match",
+		60_000, // 1 minute
+		`WorkerControllerMinSignalIntervalSyncMatchMilliseconds controls the batching interval of sync matches grouped by WCI in milliseconds (per namespace). Each batch triggers a signal.`,
+	)
+	WorkerControllerEnabledComputeProviders = dynamicconfig.NewNamespaceTypedSetting(
+		"workercontroller.compute_providers.enabled",
+		[]string(nil),
+		`WorkerControllerEnabledComputeProviders defines the list of compute providers enabled for use. Namespace-scoped: a namespace-level value overrides the cell-wide (global) value, and namespaces without an entry inherit the cell value.`,
+	)
+)
+
+// Global settings.
+var (
 	WorkerControllerAWSIntermediaryRoles = dynamicconfig.NewGlobalTypedSetting(
 		"workercontroller.compute_providers.aws.intermediary_roles",
 		[][]AWSIAMRoleRequest(nil),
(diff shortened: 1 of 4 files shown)


wci/client/config_test.go
02strich, line 1: does this test anything specific to WCI or just dynamicconfig overall?

wci/workflow/compute_provider/registry.go
```diff
@@ -69,8 +69,8 @@ func RegisterComputeProvider(providerType iface.ComputeProviderType, ctor Comput
 	}
 }
 
-func GetComputeProvider(ctx context.Context, providerType iface.ComputeProviderType, dc *dynamicconfig.Collection) (ComputeProvider, error) {
-	enabledProviders := client.WorkerControllerEnabledComputeProviders.Get(dc)()
+func GetComputeProvider(ctx context.Context, providerType iface.ComputeProviderType, namespace string, dc *dynamicconfig.Collection) (ComputeProvider, error) {
```
02strich:
```suggestion
func GetComputeProvider(ctx context.Context, providerType iface.ComputeProviderType, namespaceName string, dc *dynamicconfig.Collection) (ComputeProvider, error) {
```

wci/client/config_test.go
gcristea-temporal: This is WCI specific.

02strich: sorry, what part is?

gcristea-temporal: From [this Slack message](https://temporaltechnologies.slack.com/archives/C0A9CUY2PHA/p1782845810442869):

**Cell Config (AWS cells only):**
```
{
  "workercontroller.compute_providers.enabled": [{"value": ["aws-lambda"]}]
}
```
**Namespace Config — Cloud Run only:**
```
{
  "workercontroller.compute_providers.enabled": [{"value": ["gcp-cloud-run"]}]
}
```

**Namespace Config — Lambda only:**
No entry needed — inherits from cell DCO

wci/client/config.go
```diff
@@ -34,6 +35,25 @@ var (
...
+		"workercontroller.hook.min_signal_interval_no_sync_match",
+		500,
+		`WorkerControllerMinSignalIntervalNoSyncMatchMilliseconds controls the batching interval of no-sync matches grouped by WCI in milliseconds (per namespace). Each batch triggers a signal. `,
+	)
+	WorkerControllerMinSignalIntervalSyncMatchMilliseconds = dynamicconfig.NewNamespaceIntSetting(
+		"workercontroller.hook.min_signal_interval_sync_match",
+		60_000, // 1 minute
+		`WorkerControllerMinSignalIntervalSyncMatchMilliseconds controls the batching interval of sync matches grouped by WCI in milliseconds (per namespace). Each batch triggers a signal.`,
+	)
+	WorkerControllerEnabledComputeProviders = dynamicconfig.NewNamespaceTypedSetting(
```
ericbriffa, line 48: is lambda (and future providers) staying in global, or being removed once in public preview? 

I'm curious of the situation where the override to enable one provider disables another, can that happen?

wci/workflow/compute_provider/registry.go
gcristea-temporal: Fixed, thank you.

02strich (approved):

wci/client/config.go
gcristea-temporal: For now they have to stay in the dynamic config. I think it's a longer conversation on whether we can set the values for specific providers default to `true` in the code (changing the default value may break self-hosted.)

This change allows to override at the namespace level the value that is set at global/cell level. The namespace level value will replace any cell/global value. 

In the case we want both we'll have to use something like
```
{
  "workercontroller.compute_providers.enabled": [
    {
      "value": [
        "aws-lambda"
      ]
    },
    {
      "value": [
        "gcp-cloud-run"
      ]
    }
  ]
}
```

ericbriffa: Got it, thanks. That's a potential footgun, but the likelihood is low and this is a runbooked process so we ensure the right checks are there. 

The other use case for edits is when we go from prerelease to public preview for a given provider we have to "clean out" the namespaces of the provider that no longer needs namespace config (and revert to a global or implicit enable).  That feels harder with this vs decoupling providers into their own bools. Any concerns there?

ericbriffa (approved):

wci/client/config.go
gcristea-temporal: There is a definition for `CELL_PROVIDER` in the `dynamic_config.proto` (saas API). However I don't think that it's implemented (according to [this runbook](https://runbooks.tmprl-internal.cloud/cloud/cloud-foundation/dynamic-config/architecture#scopes)).

PR #77, opened 2026-07-09
use proto field path for Cloud Run scale update mask

gcristea-temporal:
## What was changed

Use the proto field path `scaling.manual_instance_count` and build the mask via `fieldmaskpb.New` so an invalid path panics at load instead of misbehaving silently at runtime.

## Why?

`UpdateWorkerSetSize` sent the scale update mask as `scaling.manualInstanceCount` (camelCase JSON name). 

Cloud Run silently ignores an unknown mask path and  applies the sparse WorkerPool wholesale, resetting the pool's runtime  serviceAccount to the project default compute SA — which surfaces as an `iam.serviceAccounts.actAs` PermissionDenied on the deployer.                                                                                                          

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/wci/workflow/compute_provider/gcp_cloudrun.go
+++ b/wci/workflow/compute_provider/gcp_cloudrun.go
@@ -24,6 +24,21 @@ const (
 	configGCPCloudRunServiceAccount = "service_account"
 )
 
+// scalingManualInstanceCountMask restricts a WorkerPool update to just the manual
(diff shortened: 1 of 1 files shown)


02strich:
Is this a real problem given it is a constant we are dealing with?

02strich (approved):
After chatting with the author, it is a real thing b/c Google docs are buggy and so figuring out the right constant is unreliable

PR #78, opened 2026-07-09
Update comments, fix race condition in tests

justinschoeff:
## What was changed

This updates the multiple worker deployment version integration test to
fix a race condition hit in github actions. Previously, the test would
create both workflows and then create a worker before waiting for the
invoke. However, WCI will only trigger the scale-up/invoke if there are
no pollers. So when running locally, WCI matching happens prior to
worker spin up, but in action fleet with slower hosts, it may run
slower.

This also cleans up some of the comments around CREATE status and some
descriptive strings being used in the tests.

## Why?

Race condition in test caused github action to faill sometimes. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/integration/wci_test.go
+++ b/tests/integration/wci_test.go
@@ -563,7 +563,7 @@ func TestWCIVersionInactiveAfterInvoke(t *testing.T) {
 	require.NoError(t, err)
 
 	// Before any worker has registered a task queue against the version, its
-	// status should be created.
+	// status should have status CREATED.
 	descResp, err := cli.WorkflowService().DescribeWorkerDeploymentVersion(ctx,
 		&workflowservice.DescribeWorkerDeploymentVersionRequest{
 			Namespace:         namespace,
@@ -572,7 +572,7 @@ func TestWCIVersionInactiveAfterInvoke(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CREATED,
 		descResp.GetWorkerDeploymentVersionInfo().GetStatus(),
-		"version should be CREATED before any task queue is registered")
+		"version should have status CREATED before any task queue is registered")
 
 	// WCI validates the spec then invokes workers to register task queues.
 	waitForInvoke(t, events, 60*time.Second, "register-task-queues invoke")
@@ -617,10 +617,9 @@ func TestWCIMultipleVersionsInvokeWithPinnedWorkflows(t *testing.T) {
 		})
 	require.NoError(t, err)
 
-
 	buildID1 := uuid.NewString()
 	buildID2 := uuid.NewString()
-	taskQueue := "inactive-tq" + deploymentName
+	taskQueue := "pinned-wflws-tq-" + deploymentName
 
 	version1 := &deploymentpb.WorkerDeploymentVersion{
 		DeploymentName: deploymentName,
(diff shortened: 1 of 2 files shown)


tests/integration/wci_test.go
```diff
@@ -691,7 +690,6 @@ func TestWCIMultipleVersionsInvokeWithPinnedWorkflows(t *testing.T) {
 	drainEvents(t, events2)
 	w2.Stop()
 
-
 	// Submit a workflow pinned to both versions with no pollers present, creating a backlog.
```
jaypipes: I actually don't know what "a workflow pinned to both versions" means...

tests/integration/wci_test.go
```diff
@@ -728,19 +714,31 @@ func TestWCIMultipleVersionsInvokeWithPinnedWorkflows(t *testing.T) {
 	require.NoError(t, wflow1.Get(getCtx, &result))
 	require.Equal(t, "foo", result)
 
-	waitForInvoke(t, events1, 60*time.Second, "wflow1 scale-up v2 invoke")
 	requireNoEvents(t, events2)
```
jaypipes, line 719: Shouldn't this be asserting no events on `events1`?

tests/integration/wci_test.go
```diff
@@ -728,19 +714,31 @@ func TestWCIMultipleVersionsInvokeWithPinnedWorkflows(t *testing.T) {
...
+	waitForInvoke(t, events2, 60*time.Second, "wflow2 scale-up version2 invoke")
 	w2 = startVersionedWorker(t, cli, taskQueue, deploymentName, buildID2)
 	t.Cleanup(w2.Stop)
 
 	getCtx, cancel = context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 	require.NoError(t, wflow2.Get(getCtx, &result))
 	require.Equal(t, "foo", result)
-	waitForInvoke(t, events2, 60*time.Second, "wflow1 scale-up v2 invoke")
 	requireNoEvents(t, events1)
```
jaypipes, line 745: Shouldn't this be asserting no events on `events2`?

tests/integration/wci_test.go
justinschoeff: This is poorly written and makes even less sense with this change. Before it created both workflows up front each pinned to corresponding version. 

Will update this comment.

tests/integration/wci_test.go
justinschoeff: We have two versions and created a workflow pinned to v1 only. This aims to assert that there was no events sent to the compute provider for version 2. 

This drains v1 in the next line so that we can do the same check once this workflow completes and we trigger v2 below.

jaypipes: @justinschoeff OK, perhaps this should *also* be asserting no events on `events1` then?

tests/integration/wci_test.go
jaypipes: And similarly should this be asserting there are *also* no events on `events2`?

tests/integration/wci_test.go
justinschoeff: I'm not opposed, but I'm not sure what it adds. The `waitForInvoke` on 707 will ignore other events, effectively dropping them. And while i would expect it to be empty, do you know if there's a chance another Invoke comes in before line 708 starts up the worker? I'm not familiar with how aggressively it will call invoke/scale-up. Happy to add it, if it makes sense.

jaypipes: ¯\_(ツ)_/¯ 

I would say if we expect it to be empty, we should add an assertion that it's empty :)

jaypipes (approved):

PR #79, opened 2026-07-13
WDV delete protection and queue stat tests

justinschoeff:
- **WDV delete protection integration tests**
- **Add integ coverag for WDV queue stat fields**

## What was changed

Added integ tests covering more test cases for worker deployment versions.

## Why?

Add coverage for queue stats and deletion protection.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/integration/wci_test.go
+++ b/tests/integration/wci_test.go
@@ -12,6 +12,7 @@ import (
 	deploymentpb "go.temporal.io/api/deployment/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
+	taskqueuepb "go.temporal.io/api/taskqueue/v1"
 	workflowservice "go.temporal.io/api/workflowservice/v1"
 	computeprovider "go.temporal.io/auto-scaled-workers/wci/workflow/compute_provider"
 	sdkclient "go.temporal.io/sdk/client"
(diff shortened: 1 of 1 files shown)


jaypipes (commented):
@justinschoeff @ericbriffa @smuneebahmad are these tests actually related to WCI? While they are perfectly valid functional tests for delete protection of WDVs, I don't see how they are specifically related to WCI. Rather, they seem to be a repeat of the [existing upstream tests of WDV deletion](https://github.com/temporalio/temporal/blob/8224a5375112079ad905c4ea829420306431462c/tests/worker_deployment_version_test.go#L567-L746).

Is this just to check off a line in a testing spreadsheet or are these tests supposed to be validating something specific to WCI/serverless?

justinschoeff:
@jaypipes  Yes, we are working through a list of desired test cases. I had similar concerns on overlap. I don't know that these are providing a meaningful difference from the upstream ones in the sense that we need to test various protection mechanisms. However, the tests upstream do not have [WCI enabled afaict](https://github.com/temporalio/temporal/blob/8224a5375112079ad905c4ea829420306431462c/tests/worker_deployment_version_test.go#L73-L95), while we explicitly set it in [our test env](https://github.com/temporalio/temporal-auto-scaled-workers/blob/main/tests/integration/env.go#L19). The create, delete and describe wflow do interact with WCI today. So i kinda view these as ensuring the existing WDV semantics when WCI is enabled.

jaypipes (approved):
I'm cool adding these even if they do overlap a bit with stuff in OSS server. Thanks @justinschoeff !

PR #80, opened 2026-07-14
Establish GCP global SA as impersonation base, not delegate.

gcristea-temporal:
## What was changed

* Impersonates `delegates[0]` directly as the chain base, then delegate to the customer target service account. 
* Adds tests via an impersonation seam.
* Refactors unit tests to use `assert`/`require`
* Fixed field mask used to update the instance count for Cloud Run worker pools.

## Why?
The Cloud Run provider flattened the whole chain into Delegates off the ambient ADC, which diverges from the established design. 

## Checklist

1. Closes 

2. How was this tested:

* Unit tests
* Manual end-to-end testing.

3. Any docs updates needed?

### DIFF
--- a/wci/client/config.go
+++ b/wci/client/config.go
@@ -64,6 +64,15 @@ var (
 		[][]GCPIAMServiceAccountRequest(nil),
 		`WorkerControllerGCPIntermediaryServiceAccounts defines the service account chaining to assume before assuming the per-provider role.`,
 	)
+	WorkerControllerGCPFirstDelegateAsBase = dynamicconfig.NewGlobalBoolSetting(
+		"workercontroller.compute_providers.gcp.first_delegate_as_base",
+		false,
+		`WorkerControllerGCPFirstDelegateAsBase controls how the first entry of the resolved GCP impersonation chain is treated. `+
+			`When true, the first delegate is the identity the pod's ambient Workload Identity can *directly* impersonate `+
+			`(roles/iam.workloadIdentityUser -> getAccessToken): it is used as the base credential of the chain, and the remaining `+
+			`delegates are passed as token-creator delegates to the target. When false, the entire chain is passed as token-creator `+
+			`delegates starting from the ambient ADC (default), which requires the ambient identity to hold implicitDelegation on the first delegate.`,
+	)
 	WorkerControllerEnabledScalingAlgorithms = dynamicconfig.NewGlobalTypedSetting(
 		"workercontroller.scaling_algorithms.enabled",
 		[]string(nil),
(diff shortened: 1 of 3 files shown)


wci/workflow/compute_provider/gcp_cloudrun.go
```diff
@@ -109,11 +130,32 @@ func (p *gcpCloudRunComputeProvider) buildClientAndParams(ctx context.Context, r
...
 		}
 
-		ts, err := impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
+		scopes := []string{"https://www.googleapis.com/auth/cloud-platform"}
+
+		// delegates[0] is the pool's ambient identity can *directly*
+		// impersonate (workloadIdentityUser → getAccessToken).
+		// It must be the base of the chain, not a Delegates entry: the ambient SA
+		// holds no implicitDelegation through it. Remaining entries are genuine
+		// delegates to the customer target SA.
```
02strich: I am sorry but I don't understand what this means. It seems like something is special about the first entry which seems less than ideal on the surface but maybe it's fine?

gcristea-temporal: The WCI service account has `roles/iam.workloadIdentityUser` on the global service account (the first element in the list of delegates.) 

That is why the first hop has to be done separately.

02strich: Does everybody need to do it this way? If not can we make this configurable in the dynamic config?

gcristea-temporal: The outline above only applies to Cloud and is the most complex case the code supports. It also supports cases where 
* only the first part (WIF) is used
* direct impersonation

02strich: then let's make this explicit as a setting in the DC entry that it needs this dance. I don't like hard coding current behavior of one consumer into the codebase

gcristea-temporal: I added a setting in the (global) dynamic config controlling this functionality.

wci/client/config.go
```diff
@@ -64,6 +64,15 @@ var (
 		[][]GCPIAMServiceAccountRequest(nil),
 		`WorkerControllerGCPIntermediaryServiceAccounts defines the service account chaining to assume before assuming the per-provider role.`,
 	)
+	WorkerControllerGCPFirstDelegateAsBase = dynamicconfig.NewGlobalBoolSetting(
```
02strich, line 67: any reason for not embedding this into GCPIAMServiceAccountRequest as a field?

02strich (approved):

PR #82, opened 2026-07-14
Add Worker deployment integ coverage

justinschoeff:
This adds worker deployment test coverage, ensuring consistent worker
deployment state and interactions.

## What was changed
 New tests covering worker deployment API calls.

## Why?
 WCI is directly integrated in worker deployment versions and their parent worker deployments. So adding coverage in this package to ensure the WCI updates don't break expected semantics for worker deployments.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/integration/worker_deployment_test.go
+++ b/tests/integration/worker_deployment_test.go
@@ -1,0 +1,347 @@
+package integration
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	deploymentpb "go.temporal.io/api/deployment/v1"
+	"go.temporal.io/api/serviceerror"
+	workflowservice "go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/server/tests/testcore"
+)
+
+func TestWCICreateWorkerDeploymentSuccess(t *testing.T) {
+	env := createWCITestEnv(t)
+	ctx := env.Context()
+	cli := env.SdkClient()
+
+	namespace := env.Namespace().String()
+	deploymentName := uuid.NewString()
+
+	resp, err := cli.WorkflowService().CreateWorkerDeployment(ctx,
+		&workflowservice.CreateWorkerDeploymentRequest{
+			Namespace:      namespace,
+			DeploymentName: deploymentName,
+			Identity:       "test-identity",
+			RequestId:      uuid.NewString(),
+		})
+	require.NoError(t, err)
+	require.NotEmpty(t, resp.GetConflictToken())
+
+	descResp, err := cli.WorkflowService().DescribeWorkerDeployment(ctx,
+		&workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      namespace,
+			DeploymentName: deploymentName,
+		})
+	require.NoError(t, err)
+	require.Equal(t, deploymentName, descResp.GetWorkerDeploymentInfo().GetName())
+	require.NotNil(t, descResp.GetWorkerDeploymentInfo().GetCreateTime())
+	require.Empty(t, descResp.GetWorkerDeploymentInfo().GetVersionSummaries())
+}
+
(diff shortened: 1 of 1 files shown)


tests/integration/worker_deployment_test.go
```diff
@@ -0,0 +1,374 @@
...
+	createWorkerDeployment(t, ctx, cli, namespace, deploymentName)
+
+	resp, err := cli.WorkflowService().DescribeWorkerDeployment(ctx,
+		&workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      namespace,
+			DeploymentName: deploymentName,
+		})
+	require.NoError(t, err)
+	require.NotNil(t, resp.GetWorkerDeploymentInfo())
+	require.NotEmpty(t, resp.GetConflictToken())
```
mani-j9: I think this one can be skipped in favor of TestWCICreateWorkerDeploymentSuccess covering the assert on conflictToken

tests/integration/worker_deployment_test.go
```diff
@@ -0,0 +1,374 @@
...
+	cli := env.SdkClient()
+
+	_, err := cli.WorkflowService().DescribeWorkerDeployment(ctx,
+		&workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      env.Namespace().String(),
+			DeploymentName: uuid.NewString(),
+		})
+	require.Error(t, err)
+	var notFound *serviceerror.NotFound
+	require.ErrorAs(t, err, &notFound,
```
mani-j9, line 135: nit: the trailing ,` got: %v", `err can be removed which is part of the o/p from ErrorAs.

tests/integration/worker_deployment_test.go
```diff
@@ -0,0 +1,374 @@
...
+	ctx := env.Context()
+	cli := env.SdkClient()
+
+	namespace := env.Namespace().String()
+	deploymentName := uuid.NewString()
+	createWorkerDeployment(t, ctx, cli, namespace, deploymentName)
+
+	buildID1 := uuid.NewString()
+	buildID2 := uuid.NewString()
+	createVersion(t, ctx, cli, namespace, deploymentName, buildID1)
```
mani-j9: i was thinking about using a similar pattern of passing ctx, cli and namespace to the helpers that I have introduced in my [PR](https://github.com/temporalio/temporal-auto-scaled-workers/pull/81/changes) but eventually landed at the approach of passing just the env var and letting helpers do the 3 var set up. mainly to keep it to one argument instead of three; but i'm open to changing those to keep it consistent with this code if the team prefers this. any thoughts?

tests/integration/worker_deployment_test.go
justinschoeff: fair, will merge into success.

tests/integration/worker_deployment_test.go
justinschoeff: Yeah. Changed to already exists to see. For reference, the message and type are output in the `Error:`, the `%v` is repeating that.
```
        	Error Trace:	/Users/justinschoeff/workplace/serverless/temporal-auto-scaled-workers/tests/integration/worker_deployment_test.go:137
        	Error:      	Should be in error chain:
        	            	expected: *serviceerror.AlreadyExists
        	            	in chain: "no Worker Deployment found with name '5befad92-6772-4bf9-a6c9-b011e68a1cc2'; does your Worker Deployment have pollers?" (*serviceerror.NotFound)
        	Test:       	TestWCIDescribeWorkerDeploymentNotFound
        	Messages:   	describing a missing deployment should return NotFound, got: no Worker Deployment found with name '5befad92-6772-4bf9-a6c9-b011e68a1cc2'; does your Worker Dep
```

tests/integration/worker_deployment_test.go
justinschoeff: I can update, i almost did what you had, but left it at mutliple args. I don't have a strong opinion.

mani-j9 (approved):

PR #86, opened 2026-07-17
Initial impl. adding compute provider to metrics

seanbollin:
## What was changed
Added compute_provider to all metrics that are in the context of a scaling group (since compute provider is defined per scaling group). 

Screenshot 2026-07-21 at 10 26 01 AM

worker_controller_instance_operations
worker_controller_instance_scale_up_count
worker_controller_instance_deferred_scaling_decision_count
worker_controller_instance_activities
worker_controller_instance_scale_up_throttled_count
worker_controller_instance_deferred_scaling_decision_metrics_pull_failed_count

## Why?
So that we can tell which compute_provider is being used on the metrics dashboard.

## Checklist

1. Closes (https://temporalio.atlassian.net/browse/COM-131)

2. How was this tested:
Verified metric emission on a test cell

3. Any docs updates needed?
Will update dashboards


wci/workflow/workflow.go
```diff
@@ -505,7 +512,7 @@ func (d *WorkflowRunner) periodicValidateSpec(ctx workflow.Context) {
 			Spec:           d.State.Spec,
 		},
 	).Get(ctx, nil); err != nil {
-		d.recordOperation(wcimetrics.OperationTypeValidateSpec, wcimetrics.ErrorTypeActivityError, classifyActivityErrorType(err), wcimetrics.SkippedReasonNone)
+		d.recordOperation(wcimetrics.OperationTypeValidateSpec, "", wcimetrics.ErrorTypeActivityError, classifyActivityErrorType(err), wcimetrics.SkippedReasonNone)
```
mani-j9: nit (feel free to ignore): since computeProviderTagValue("") already maps "" → "none", would it be worth passing a typed none sentinel at these call sites instead of a bare ""?

wci/metrics/metric_defs.go
```diff
@@ -55,6 +56,13 @@ const (
 	ActivityErrorTypeTagName   = "activity_error_type"
 )
 
+// ComputeProviderNone is the sentinel compute_provider tag value used when a
+// metric is not emitted in the context of a single scaling group (so no one
+// provider applies). Keeps the tag a visible, queryable value rather than an
+// empty string that some exporters drop — matching the ErrorTypeNone /
+// SkippedReasonNone / ActivityErrorTypeNone convention.
```
02strich:
```suggestion
// ComputeProviderNone is the sentinel compute_provider tag value used when a
// metric is not emitted in the context of a single scaling group (so no one
// provider applies).
```

wci/workflow/compute_provider/registry.go
```diff
@@ -21,9 +21,10 @@ type (
 	// activity request type. Compute providers receive it to act on behalf of a
 	// specific namespace/deployment (e.g. resolving the GCP impersonation chain).
 	RequestContext struct {
-		NamespaceName     string `json:"namespace_name"`
-		DeploymentName    string `json:"deployment_name"`
-		DeploymentBuildID string `json:"deployment_build_id"`
+		NamespaceName     string                    `json:"namespace_name"`
+		DeploymentName    string                    `json:"deployment_name"`
+		DeploymentBuildID string                    `json:"deployment_build_id"`
+		ComputeProvider   iface.ComputeProviderType `json:"compute_provider,omitempty"`
```
02strich: the compute provider already knows it's own type so this seems not needed

02strich (changes requested):

ericbriffa:
Thanks for this. Are these metrics going into both prometheus and chronicle?   

For querying post-change, would we be doing something like `compute_provider IN ('lambda', 'none')` to identify lambda?

seanbollin:
> Thanks for this. Are these metrics going into both prometheus and chronicle?
> 
> For querying post-change, would we be doing something like `compute_provider IN ('lambda', 'none')` to identify lambda?

That is certainly the idea. I've seen them show up in prometheus but not clickhouse. Not sure if there's anything else to do to get them into clickhouse - I didn't think so but asking about it.

Yes something like that for computer_provider, but if you want to know it's lambda I wouldn't check against none, just aws-lambda ..

02strich (approved):

mani-j9 (approved):
🔥

PR #89, opened 2026-07-20
Add scaling action processing latency metric

seanbollin:
## What was changed
Added a scaling action processing latency to track time from work received to scaling action

Screenshot 2026-07-22 at 2 21 37 PM
Screenshot 2026-07-22 at 2 22 10 PM

## Why?
Track on the SLO dashboard

## Checklist

1. Closes https://temporalio.atlassian.net/browse/COM-223

2. How was this tested:

Deployed to test cell and verified metric emission

3. Any docs updates needed?

### DIFF
--- a/wci/workflow/workflow.go
+++ b/wci/workflow/workflow.go
@@ -484,7 +484,7 @@ func (d *WorkflowRunner) pullStatsAndUpdate(ctx workflow.Context) time.Duration 
 		if resp.UpdatedScalingStatus != nil {
 			d.State.ScalingStatus = resp.UpdatedScalingStatus
 		}
-		d.handleActions(ctx, resp.Actions, nil, actionLatencyOrigin{start: pollStart.Add(-resp.BacklogAge), path: wcimetrics.PathStats})
+		d.handleActions(ctx, resp.Actions, nil, scalingActionLatencyOrigin{start: pollStart.Add(-resp.BacklogAge), path: wcimetrics.PathPullStats})
 
 		return time.Duration(resp.NextPollSeconds) * time.Second
 	}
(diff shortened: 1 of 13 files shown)


wci/workflow/workflow.go
```diff
@@ -489,7 +491,7 @@ func (d *WorkflowRunner) pullStatsAndUpdate(ctx workflow.Context) time.Duration
 		if resp.UpdatedScalingStatus != nil {
 			d.State.ScalingStatus = resp.UpdatedScalingStatus
 		}
-		d.handleActions(ctx, resp.Actions, nil)
+		d.handleActions(ctx, resp.Actions, nil, scalingActionLatencyOrigin{start: pollStart.Add(-resp.BacklogAge), path: wcimetrics.PathPullStats})
```
02strich: why is this including backlog age? that seems confusing

wci/workflow/workflow.go
```diff
@@ -566,10 +569,17 @@ func (d *WorkflowRunner) handleNoSyncMatchSignal(ctx workflow.Context, req *ifac
 		if resp.UpdatedScalingStatus != nil {
 			d.State.ScalingStatus = resp.UpdatedScalingStatus
 		}
-		d.handleActions(ctx, resp.Actions, req)
+		d.handleActions(ctx, resp.Actions, req, scalingActionLatencyOrigin{start: signalReceiptTime, path: wcimetrics.PathTaskAdd})
```
02strich: signal receipt time is not the start of the process, that would be in matching as there can be delays here. Also did you validate that `workflow.Now` really returns the right time?

seanbollin: @02strich @ericbriffa before I revise, I want to align on what this metric actually measures, since a couple of the review comments touch the core definition. Here's where I landed after talking with Eric:

Two separate metrics:

1. scaling_action_latency (the SLO metric) - WCI processing latency (no matching service consideration), measured the same way on both paths:

- task_add: WCI signal receipt -> scaling action
- pull_stats: WCI poll start -> scaling action (excluding backlog age)

2. backlog age - kept as a separate diagnostic metric (how long work waited before we acted, stats path only). Different metric name so it never gets folded into the combined query

wci/workflow/workflow.go
seanbollin: I will move backlog age into a separate metric so the SLO metric no longer mixes two definitions of start

wci/workflow/workflow.go
seanbollin: Yes there is latency in the matching service before WCI sees the work - this proposal deliberately scopes the combined SLO to WCI processing latency (from when WCI starts handling) - is that scoping OK?

seanbollin: Yes I did test with workflow.Now - It is quantized to the workflow task start time so it is imprecise but replay-safe. Alternatively we could go with a SideEffect to measure time.Now but that comes at the cost of another workflow history event - do you have any input on which would be better?

02strich: I think the two metrics make sense, but would rename the first one to be explicitly about the in-wci processing latency.

If you doublechecked the worflow.Now I am good with using it

02strich (approved):
While I think the scaling action latency should have another name, already approving it as we can adjust that

PR #90, opened 2026-07-20
Support worker-set integration tests

justinschoeff:
This adds support for integration coverage of the worker-set launch
strategy. This refactors the existing test_invoke and
test_invoke_observer to suport other compute providers instead of
simply invoke strategy.

## What was changed

This adds support for integration coverage of the worker-set launch
strategy. This refactors the existing test_invoke and
test_invoke_observer to suport other compute providers instead of
simply invoke strategy.

## Why?

Coverage of upcoming worker set compute provider.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/integration/worker_set_test.go
+++ b/tests/integration/worker_set_test.go
@@ -1,0 +1,502 @@
+package integration
+
+import (
+	"context"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	computepb "go.temporal.io/api/compute/v1"
+	deploymentpb "go.temporal.io/api/deployment/v1"
+	"go.temporal.io/api/serviceerror"
+	workflowservice "go.temporal.io/api/workflowservice/v1"
+	wciworkflow "go.temporal.io/auto-scaled-workers/wci/workflow"
+	computeprovider "go.temporal.io/auto-scaled-workers/wci/workflow/compute_provider"
+	sdkclient "go.temporal.io/sdk/client"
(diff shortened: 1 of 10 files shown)


tests/integration/worker_set_test.go
```diff
@@ -0,0 +1,502 @@
...
+	ctx := env.Context()
+	cli := env.SdkClient()
+
+	namespace := env.Namespace().String()
+	deploymentName := uuid.NewString()
+	buildID := uuid.NewString()
+	taskQueue := "workerset-scaleup2-tq-" + deploymentName
+	version := &deploymentpb.WorkerDeploymentVersion{DeploymentName: deploymentName, BuildId: buildID}
+
+	spy := &invokeSpy{events: make(chan string, 16)}
```
tylercunnion, line 163: I wonder if there's an opportunity to pull this into a helper function, we repeat this pattern in pretty much every test

tylercunnion (approved):
Nice, I like the work here to tighten up the timing as well. Will that take effect in the Invoke-based tests as well or do we need to do anything to make that happen?

tests/integration/worker_set_test.go
justinschoeff: Agreed. @mani-j9 added some helpers for creating versioned workers, i was hoping to refactor the existing tests to match. I'll look at including the spy as well.

justinschoeff:
@tylercunnion I think the existing tests will get benefit from the added dynamic config for no sync match interval. I'll take a look at using the min/maxinterval in existing tests.

PR #91, opened 2026-07-22
Add worker-set scaling metrics

mani-j9:
Adds worker-set scale up/down and target-size metrics: reuses `worker_controller_instance_scale_up_count` for the worker-set path, adds `worker_controller_instance_scale_down_count` and `worker_controller_instance_target_worker_count`.

https://temporalio.atlassian.net/browse/COM-43

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wci/workflow/scaling_algorithm/registry.go
```diff
@@ -22,6 +22,9 @@ type (
 		ScalingGroupKey string     `json:"scaling_group_key"`
 		Action          ActionType `json:"action"`
 		Count           *int32     `json:"count,omitempty"`
+		// PreviousCount is the worker-set size before this change, set only for
+		// ActionTypeUpdateWorkerSetSize (nil otherwise) to help derive the scale direction downstream.
+		PreviousCount *int32 `json:"previous_count,omitempty"`
```
02strich: this seems like a hack - can we avoid adding this here?

02strich (commented):
Given we are tracking the target count, do we really have to track scale up vs scale down counts?

mani-j9:
> Given we are tracking the target count, do we really have to track scale up vs scale down counts?

Good point, target_worker_count does already capture scaling direction (increase vs decrease) and magnitude. I've removed the other metrics in the latest rev.

02strich (approved):

jaypipes (approved):
👍 thanks @mani-j9 makes sense to me.

PR #94, opened 2026-07-23
Refactoring metrics helpers in the workflow

seanbollin:
## Why?

A lot of redundant and verbose code for emitting metrics

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


02strich (approved):
The private type names could be a bit more explicit to metrics, but otherwise looks like a nice improvement.

wci/workflow/metrics_recorders.go
```diff
@@ -0,0 +1,170 @@
+package workflow
```
seanbollin, line 1: I think I'll add a unit test to this that lists all metrics and what labels they should have

jaypipes: @seanbollin That kind of a unit test generally just provides friction when evolving the set of metrics, in my experience. Better to skip that kind of a unit test (that really is just going to be a repeat of the const/var definitions in `wci/metrics/metrics_defs.go`) and instead focus on tests that verify the *value* of a particular metric has been updated in an expected way during a single execution of some unit of work -- for example, that a particular counter is incremented or that a duration is within an acceptable range.

PR #98, opened 2026-07-24
COM-231: Support worker set launch strategy in InvokeWorkersToRegisterTaskQueues 

gcristea-temporal:
## What was changed
Extended `InvokeWorkersToRegisterTaskQueues` to support `computeprovider.LaunchStrategyWorkerSet`.
## Why?

WCI was not scaling up for Cloud Run, resulting in the task queue for the worker deployment version not being created.

## Checklist

1. Closes 
[COM-231](https://temporalio.atlassian.net/browse/COM-231)

[COM-231]: https://temporalio.atlassian.net/browse/COM-231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wci/workflow/activities.go
+++ b/wci/workflow/activities.go
@@ -36,6 +36,11 @@ const (
 var (
 	minPollInterval = 30 * time.Second
 	maxPollInterval = 5 * time.Minute
+
+	// registerTaskQueueWorkerSetSize is the instance count a worker-set provider is
+	// scaled to so a single worker comes up and registers the task queue. The scaling
+	// algorithm resizes the set from there.
+	registerTaskQueueWorkerSetSize int32 = 1
 )
 
 type (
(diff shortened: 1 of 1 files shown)


justinschoeff (approved):
Do we want to add tests in a follow-on, I don't see coverage in activities_test currently? We do have coverage for it in integ tests though.

wci/workflow/activities.go
```diff
@@ -36,6 +36,11 @@ const (
 var (
 	minPollInterval = 30 * time.Second
 	maxPollInterval = 5 * time.Minute
+
+	// registerTaskQueueWorkerSetSize is the instance count a worker-set provider is
+	// scaled to so a single worker comes up and registers the task queue. The scaling
+	// algorithm resizes the set from there.
+	registerTaskQueueWorkerSetSize int32 = 1
```
mani-j9, line 43: should this be seeding to initial_count instead of 1?

gcristea-temporal: Initial count may be `0`

mani-j9: ah you are correct, that is possible. I'm thinking through a scenario where the update to 1 goes unnoticed by the scaling logic when it reads the actual state as initial_count instead of 1 and skips performing a scaling action. 

say a customer set it to initial_count, and we primed it down to 1. and the scaling algo thinks desired == current == initial_count and skips scaling action here https://github.com/temporalio/temporal-auto-scaled-workers/blob/43fc6ebc05afe3b1aea05af9037eb0cd79944fcf/wci/workflow/scaling_algorithm/rate_based.go#L559. i may be overthinking but is that a concern?

wci/workflow/activities.go
```diff
@@ -261,17 +266,25 @@ func (a *Activities) InvokeWorkersToRegisterTaskQueues(ctx context.Context, req
...
 			if err := provider.InvokeWorker(ctx, req.RequestContext, config); err != nil {
 				recordError(wcimetrics.ErrorTypeComputeProviderFailed)
 				return temporal.NewApplicationErrorWithCause(fmt.Sprintf("%s: %s", k, err.Error()), "InvokeWorkerFailed", err)
 			}
+		case computeprovider.LaunchStrategyWorkerSet:
+			// Scale the set up so a worker comes up and registers the task queue.
+			if err := provider.UpdateWorkerSetSize(ctx, req.RequestContext, config, registerTaskQueueWorkerSetSize); err != nil {
+				recordError(wcimetrics.ErrorTypeComputeProviderFailed)
+				return temporal.NewApplicationErrorWithCause(fmt.Sprintf("%s: %s", k, err.Error()), "InvokeWorkerFailed", err)
+			}
```
02strich, line 287: with this approach, the scaling algorithm will misunderstand the current worker set size (and also this can break the existing worker pool if it already processes things).

PR #100, opened 2026-07-26
COM-231: Skip task-queue registration when the queue already exists

gcristea-temporal:
Only update worker size set when task queues don't exists. When updating the worker size set take into account min, initial and current number of instances (and write-back to the scaling algorithm the value.)

## What was changed
This change gates registration on actual queue existence and sizes it off the algorithm's planned count:
- Query `DescribeWorkerDeploymentVersion.VersionTaskQueues` and skip any group whose task types are all already registered. The association is durable for the version's lifetime, so registered groups need no invocation and their live worker sets are left untouched. Fail open (invoke on describe error) so a transient failure can't block an update.
- For groups that **do need** registration (at least one queue type missing), sets the worker-set resize to `max(1, PlannedCount)` so it honors min/initial (instead of clobbering to 1). Then write the size back into the algorithm's status so the next metrics poll sees the real count.

## Why?

The call to`InvokeWorkersToRegisterTaskQueues` unconditionally resized every worker-set group to `1` on each spec update, which desynced the rate-based algorithm's model (`stateRateBasedWorkerCount`, whose unset fallback is `initial_count`) from provider reality: it could shrink a live set, under-provision below the configured floor, or strand/orphan a registration worker at `1` when the algorithm believed instance count to be `0`.
## Checklist

1. Closes 
- [COM-231](https://temporalio.atlassian.net/browse/COM-231)
2. How was this tested:

- Unit tests
- Integration tests
- Manual testing, details in [COM-231](https://temporalio.atlassian.net/browse/COM-231)
3. Any docs updates needed?
- None

[COM-231]: https://temporalio.atlassian.net/browse/COM-231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wci/workflow/activities.go
+++ b/wci/workflow/activities.go
@@ -2,6 +2,7 @@ package workflow
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"maps"
 	"slices"
@@ -9,7 +10,9 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
 	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/serviceerror"
 	workflowservice "go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/auto-scaled-workers/wci/client"
 	wcimetrics "go.temporal.io/auto-scaled-workers/wci/metrics"
@@ -36,11 +39,6 @@ const (
 var (
 	minPollInterval = 30 * time.Second
 	maxPollInterval = 5 * time.Minute
-
-	// registerTaskQueueWorkerSetSize is the instance count a worker-set provider is
-	// scaled to so a single worker comes up and registers the task queue. The scaling
-	// algorithm resizes the set from there.
-	registerTaskQueueWorkerSetSize int32 = 1
 )
 
 type (
@@ -121,6 +119,17 @@ type (
 	InvokeWorkersToRegisterTaskQueuesRequest struct {
 		RequestContext
 		iface.WorkerControllerInstanceSpec
+
+		// ScalingStatus is the per-group scaling status, used to size a worker-set
+		// registration resize to the algorithm's planned count.
+		ScalingStatus map[string]iface.ScalingAlgorithmStatus `json:"scaling_status,omitempty"`
+	}
+
+	InvokeWorkersToRegisterTaskQueuesResponse struct {
+		// UpdatedScalingStatus is the complete next-state scaling status for every group in
+		// the spec: resized groups carry their reconciled count, skipped (already-registered)
+		// groups carry their prior status forward. The workflow assigns it wholesale.
+		UpdatedScalingStatus map[string]iface.ScalingAlgorithmStatus `json:"scaling_status,omitempty"`
 	}
 )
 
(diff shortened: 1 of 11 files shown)


wci/workflow/scaling_algorithm/rate_based.go
```diff
@@ -785,6 +785,26 @@ func currentRateBasedPlannedCount(config iface.ScalingAlgorithmConfig, state ifa
...
+	return min(max(planned, minCount), maxCount)
+}
+
+func (a *scalingAlgorithmRateBased) WithPlannedCount(status iface.ScalingAlgorithmStatus, count int32) iface.ScalingAlgorithmStatus {
+	if status == nil {
+		status = iface.ScalingAlgorithmStatus{}
+	}
+	if count < 0 {
+		count = 0
+	}
```
jaypipes: Instead of this runtime check, how about making `ScalingAlgorithm.PlannedCount` return a `uint32`? that way, you get compile-time checks.

wci/workflow/scaling_algorithm/rate_based_test.go
```diff
@@ -25,6 +25,63 @@ func oldMs() int64 {
...
+	t.Run("overwrites an existing worker_count", func(t *testing.T) {
+		status := iface.ScalingAlgorithmStatus{stateRateBasedWorkerCount: int64(2)}
+		got := a.WithPlannedCount(status, 7)
+		assert.Equal(t, int32(7), a.PlannedCount(iface.ScalingAlgorithmConfig{}, got))
+	})
+
+	t.Run("clamps negative counts to zero", func(t *testing.T) {
+		got := a.WithPlannedCount(nil, -3)
+		assert.Equal(t, int32(0), a.PlannedCount(iface.ScalingAlgorithmConfig{}, got))
+	})
```
jaypipes: and this unit test wouldn't even be necessary :)

wci/workflow/activities.go
```diff
@@ -244,52 +255,81 @@ func (a *Activities) ValidateSpec(ctx context.Context, req *ValidateSpecRequest)
...
 	}
 
+	logger := activity.GetLogger(ctx)
 	metricsHandler := metricsHandler(ctx, req.RequestContext, wcimetrics.ActivityTypeInvokeWorkersToRegisterTaskQueues, noComputeProvider)
 	_, _, recordSuccess := newActivityRecorders(metricsHandler)
 
+	// A task queue type is registered once a worker has polled it under this version;
+	// the association is durable for the version's lifetime, so there's no need for a
+	// worker invocation. Fail open (empty set results in a worker invocation) so a describe hiccup
+	// cannot block an update.
```
jaypipes, line 268: I know you're trying to be resilient to transient errors here, which is A Good Thing! However, consider the case where the error returned here is a rate limiting failure. If we just continue on after receiving that, do we run the risk of exacerbating that rate limiting problem?

gcristea-temporal: Changed to not continue when receiving `ResourceExhausted` type error from `DescribeWorkerDeploymentVersion`.

wci/workflow/scaling_algorithm/registry.go
```diff
@@ -90,6 +90,18 @@ type (
...
+		// derived from persisted status (falling back to config) and clamped to the algorithm's
+		// configured bounds. It is used to size an out-of-band worker-set resize (task-queue
+		// registration) so the resize respects the algorithm's floor and initial count rather
+		// than clobbering a live set.
+		PlannedCount(config iface.ScalingAlgorithmConfig, status iface.ScalingAlgorithmStatus) int32
+
+		// WithPlannedCount returns status with the planned worker count set to n, reconciling
+		// the algorithm's model with an out-of-band resize so a later decision does not see a
+		// phantom count.
+		WithPlannedCount(status iface.ScalingAlgorithmStatus, count int32) iface.ScalingAlgorithmStatus
```
02strich: I am not a fan of this API as it is hard to scale. Why is it needed to have these as dedicated methods on the interface?

wci/workflow/workflow.go
```diff
@@ -394,25 +396,40 @@ func (d *WorkflowRunner) handleUpdateInstance(ctx workflow.Context, args *iface.
...
 			d.recordUpdate(wcimetrics.UpdateTypeUpdateInstance, wcimetrics.ErrorTypeActivityError, classifyActivityErrorType(err))
 
 			if appErr, ok := errors.AsType[*temporal.ApplicationError](err); ok {
-				if appErr.Type() == "InvalidArgument" {
+				switch appErr.Type() {
+				case "InvalidArgument":
 					return nil, serviceerror.NewInvalidArgumentf("%s", appErr.Message())
+				case "ResourceExhausted":
+					// no need to pass through the cause as the client will strip that before returning
+					return nil, serviceerror.NewResourceExhausted(enumspb.RESOURCE_EXHAUSTED_CAUSE_UNSPECIFIED, appErr.Message())
```
02strich, line 417: we are not doing this case in other settings - is that something needed everywhere?

wci/workflow/workflow.go
```diff
@@ -394,25 +396,40 @@ func (d *WorkflowRunner) handleUpdateInstance(ctx workflow.Context, args *iface.
...
 		}
 
+		// Fold the registration resize back into the algorithm's model so the next metrics
+		// poll sees the actual worker-set size rather than an incorrect count.
+		if len(regResp.UpdatedScalingStatus) > 0 && d.State.ScalingStatus == nil {
+			d.State.ScalingStatus = map[string]iface.ScalingAlgorithmStatus{}
+		}
+		for _, key := range workflow.DeterministicKeys(regResp.UpdatedScalingStatus) {
+			d.State.ScalingStatus[key] = regResp.UpdatedScalingStatus[key]
+		}
```
02strich: we don't do this more complex variant in other cases (instead using a plain assignment). Is that an intentional difference?

wci/workflow/scaling_metrics_snapshot.go
```diff
@@ -65,6 +65,49 @@ func (a *Activities) pullScalingMetricsSnapshot(ctx context.Context, namespaceNa
 	return &metricsSnapshot, nil
 }
 
+// registeredTaskQueueTypes returns the set of task queue types already registered for the
```
02strich: nit: this doesn't have anything to do with the scaling metrics snapshot, right?

02strich (changes requested):
The PlannedCount and WithPlannedCount seem to violate the encapsulation of the scaling algorithm. I would have thought that if you are going through the scaling algo (which seems reasonable) you should be able to use the no-sync match path as-is, no?

wci/workflow/scaling_algorithm/rate_based.go
gcristea-temporal: No longer applies

wci/workflow/scaling_algorithm/rate_based_test.go
gcristea-temporal: No longer applies

wci/workflow/workflow.go
gcristea-temporal: Refactored the code and added a more detailed comment why the merge is needed.

wci/workflow/scaling_metrics_snapshot.go
gcristea-temporal: Moved this login to its own file.

gcristea-temporal:
> The PlannedCount and WithPlannedCount seem to violate the encapsulation of the scaling algorithm. I would have thought that if you are going through the scaling algo (which seems reasonable) you should be able to use the no-sync match path as-is, no?

@02strich Refactored the implementation based on your suggestion. I added a new method (`RegistrationActions`) to the `ScalingAlgorithm`. That method contains the logic that was accessing the current count.

wci/workflow/scaling_algorithm/registry.go
gcristea-temporal: No longer applies

wci/workflow/activities.go
```diff
@@ -244,52 +253,105 @@ func (a *Activities) ValidateSpec(ctx context.Context, req *ValidateSpecRequest)
...
+			recordError(wcimetrics.ErrorTypeAlgorithmFailed)
+			return nil, temporal.NewApplicationErrorWithCause(fmt.Sprintf("%s: %s", k, err.Error()), "InvalidArgument", err)
+		}
+		regResp, err := algo.RegistrationActions(ctx, scalingConfig, req.ScalingStatus[k])
+		if err != nil {
+			recordError(wcimetrics.ErrorTypeAlgorithmFailed)
+			return nil, temporal.NewApplicationErrorWithCause(fmt.Sprintf("%s: %s", k, err.Error()), "AlgorithmFailed", err)
+		}
+
+		for _, action := range regResp.Actions {
```
02strich, line 330: any reason for not using `handleActions`?

wci/workflow/workflow.go
```diff
@@ -394,25 +397,43 @@ func (d *WorkflowRunner) handleUpdateInstance(ctx workflow.Context, args *iface.
...
 				return nil, serviceerror.NewFailedPreconditionf("%s", appErr.Message())
 			}
 			return nil, err
 		}
 
+		// Fold the registration resize back into the algorithm's model so the next metrics
+		// poll sees the actual worker-set size rather than an incorrect count. The activity
+		// omits groups it skipped (already registered) — so the map is a subset of the spec's
+		// groups; merge rather than replace so a skipped, still-live group keeps its status
+		// instead of being reset to initial_count on the next poll.
```
02strich: this seems like a bug in the activity - it should return the correct map and not need fixup

02strich (commented):
I like the approach, few smaller comments

wci/workflow/scaling_algorithm/registry.go
```diff
@@ -90,6 +100,14 @@ type (
...
 		// The returned Actions MUST NOT contain ActionTypeDeferredScalingDecision as it is not supported.
 		ProcessMetricsPoll(ctx context.Context, config iface.ScalingAlgorithmConfig, priorStatus iface.ScalingAlgorithmStatus, metricsSnapshot ScalingMetricsSnapshot) (*MetricsPollResponse, error)
+
+		// RegistrationActions returns the actions that bring a group's worker set up enough to
+		// register its task queues, plus the reconciled status. The algorithm owns both the size
+		// and the launch strategy: worker-set algorithms emit an UpdateWorkerSetSize sized to
+		// their floor/initial count (at least 1) and fold that count into the returned Status;
+		// invoke-only algorithms emit an InvokeWorker and leave Status untouched. The caller
+		// applies the actions and persists the Status.
+		RegistrationActions(ctx context.Context, config iface.ScalingAlgorithmConfig, status iface.ScalingAlgorithmStatus) (*RegistrationResponse, error)
```
jaypipes: I hate to be a pedant, but having an interface method called `RegistrationActions` on an interface named `ScalingAlgorithm` that describes something way bigger than just an algorithm all in a file called `registry.go` is kinda killing me :)

If we're not going to rename `ScalingAlgorithm` to `Scaler` or `WorkerController` or `WDVController`, etc, perhaps we could name this particular method something more specific like `InitialTaskQueueActions` or even `TaskQueueRegistrationActions`. As it stands, `RegistrationActions` reads like this method adds a callback function for when the ScalingAlgorithm is registered with... the registry.

gcristea-temporal: Renamed to `TaskQueueRegistrationActions`

wci/workflow/workflow.go
gcristea-temporal: Fixed.

wci/workflow/activities.go
gcristea-temporal: Mainly because the end caller for `handleUpdateInstance` is expecting a result. Rolling that into `handleActions` changes its semantics and would require significant changes.

wci/workflow/workflow.go
gcristea-temporal: I don't think this pattern is needed in other places. It returns a type of error to signal the end caller that something in the call stack is rate-limited.

jaypipes (approved):
@gcristea-temporal thanks for the rename :) Sorry for being pedantic!

02strich (approved):

PR #102, opened 2026-07-27
Gate AWS Lambda ExternalID enforcement behind require_role_and_external_id

mani-j9:
Gate the `checkExternalID` enforcement probe behind `require_role_and_external_id` so the flag governs the whole role/external-ID behavior, not just presence. Previously the probe always ran when a role + external ID were present, calling `AssumeRole` without the external ID and expecting `AccessDenied` — which a permissive local STS (e.g. LocalStack) never returns, blocking local dev validation of serverless workers.

Default stays `true`, so production is unchanged. Set `false` only on isolated local/dev clusters.

### DIFF
--- a/wci/workflow/compute_provider/aws_lambda.go
+++ b/wci/workflow/compute_provider/aws_lambda.go
@@ -73,8 +73,11 @@ func (p *awsLambdaComputeProvider) ValidateConfig(ctx context.Context, _ Request
 		return fmt.Errorf("cannot access the compute resource: %w", err)
 	}
 
-	if err := p.checkExternalID(ctx, cfg, arn); err != nil {
-		return fmt.Errorf("IAM role trust policy does not enforce ExternalID condition: %w", err)
+	// Only enforce the ExternalID trust-policy check when role/external-ID are required.
+	if p.requireRoleAndExternalID {
+		if err := p.checkExternalID(ctx, cfg, arn); err != nil {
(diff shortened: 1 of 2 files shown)


wci/workflow/compute_provider/aws_lambda.go
```diff
@@ -73,8 +73,11 @@ func (p *awsLambdaComputeProvider) ValidateConfig(ctx context.Context, _ Request
 		return fmt.Errorf("cannot access the compute resource: %w", err)
 	}
 
-	if err := p.checkExternalID(ctx, cfg, arn); err != nil {
-		return fmt.Errorf("IAM role trust policy does not enforce ExternalID condition: %w", err)
+	// Only enforce the ExternalID trust-policy check when role/external-ID are required.
+	if p.requireRoleAndExternalID {
```
02strich, line 77: is this really needed? if role or eid are empty strings, the `checkExternalID` already defaults to pass.

mani-j9: empty strings don't work either because of the cli [validation](https://github.com/temporalio/cli/blob/e219b9c32c04fcd47d498d097335958492f0b0ce/internal/temporalcli/commands.worker.deployment.go#L889-L891)

mani-j9: instead of this PR, should I relax the CLI validation instead?

02strich: yeah seems like it

PR #106, opened 2026-08-05
Queue task-add signals for sequential processing behind a workflow patch

02strich:
## What was changed
Add a versioned code path that queues incoming task-add signals into local state and processes them one at a time in the main run loop, rather than handling each signal inline. This aims to avoid the case of processing multiple signals with the bare signal handler (which led to cases of the CAN condition being starved and the workflow getting terminated).

## Why?
During testing we found a case where a constant signal load would starve the CaN condition leading to an eventual termination of the workflow. In this setting the launch of a new lambda was interleaved with the processing of the next signal so that a signal handler was always open (as two ended up happening at the same time). This change aims to fix that by disallowing the signal handler concurrency.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/wci/workflow/workflow.go
+++ b/wci/workflow/workflow.go
@@ -30,10 +30,13 @@ const (
 	RegisterTaskQueuesViaWorkersActivityTimeout  = 30 * time.Second
 
 	periodicValidationInterval = 6 * time.Hour
-	maxPendingTaskAddSignals   = 4000
 
-	taskAddSignalQueueProcessingPatch = "taskAddSignalQueueProcessing"
-	taskAddSignalQueueLimitPatch      = "taskAddSignalQueueLimit"
+	// maxPendingTaskAddSignals defines how many past signals are kept around. At the
+	// default batching of 500ms this means we keep a bit more than 30min around, which
+	// seems like a reasonable cutoff while managing workflow state size.
+	maxPendingTaskAddSignals = 4000
+
+	taskAddSignalQueueLimitPatch = "taskAddSignalQueueLimit"
 )
 
 type WorkerControllerInstanceWorkflowVersion int64
@@ -196,9 +199,8 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 		return err
 	}
 
-	useQueuedTaskAddSignals := workflow.GetVersion(ctx, taskAddSignalQueueProcessingPatch, workflow.DefaultVersion, 1) > workflow.DefaultVersion
 	d.limitPendingTaskAddSignals = workflow.GetVersion(ctx, taskAddSignalQueueLimitPatch, workflow.DefaultVersion, 1) > workflow.DefaultVersion
-	if !useQueuedTaskAddSignals {
+	if !d.limitPendingTaskAddSignals {
 		// Process the signals from the prior run (pre-CaN)
 		d.processPendingTaskAddSignals(ctx)
 	}
@@ -209,7 +211,7 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 		var req *iface.SignalTaskAddRequest
 		c.Receive(ctx, &req)
 
-		if useQueuedTaskAddSignals {
+		if d.limitPendingTaskAddSignals {
 			d.queueTaskAddSignal(req)
 		} else {
 			d.handleNoSyncMatchSignal(ctx, req)
@@ -261,13 +263,13 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 	// before we start processing, we quickly drain the task add signals into the queue to make sure
 	// they get processed in this workflow task and not further queued for the next one. This makes
 	// it easier to debug the resulting queue.
-	if useQueuedTaskAddSignals {
+	if d.limitPendingTaskAddSignals {
 		d.drainTaskAddSignalChannelToQueue()
 	}
 
 	// Keep waiting for signals, when it's time to CaN the main goroutine will exit.
 	for !d.shouldContinueAsNew(ctx) {
-		if useQueuedTaskAddSignals && d.processNextQueuedTaskAddSignal(ctx) {
+		if d.limitPendingTaskAddSignals && d.processNextQueuedTaskAddSignal(ctx) {
 			continue
 		}
 
@@ -291,7 +293,7 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 	// we pass the current state as input to the next workflow execution, resulting in a new
 	// workflow history with just two initial events. This minimizes the risk of NDE (Non-Deterministic Execution)
 	// errors during server rollbacks.
-	if !useQueuedTaskAddSignals {
+	if !d.limitPendingTaskAddSignals {
 		d.drainPendingTaskAddSignals()
 	}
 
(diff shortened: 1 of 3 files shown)


jaypipes:
@02strich got a link to a bug report or even a Jira issue that demonstrates this from your PR summary?

> During testing we found a case where a constant signal load would starve the CaN condition leading to an eventual termination of the workflow. In this setting the launch of a new lambda was interleaved with the processing of the next signal so that a signal handler was always open (as two ended up happening at the same time).

I'm having trouble visualizing the "constant signal load would starve the  CaN condition" part...

02strich:
@jaypipes sorry for the long delay, the Jira is https://temporalio.atlassian.net/browse/COM-240

The background of the issue is that the prior signal handler included running an activity, which lead to the workflow processing signals back to back without checking for the CaN condition. If the load pattern is such that between the time that activity is scheduled, and completed another signal is received the code would directly continue processing the new signal instead of checking the CaN condition. Leading to a loop until the history runs out of space. That is not how this was intended to work: the idea was that after each signal the CaN condition is checked. This PR fixes that case.

wci/workflow/workflow.go
```diff
@@ -30,6 +30,10 @@ const (
 	RegisterTaskQueuesViaWorkersActivityTimeout  = 30 * time.Second
 
 	periodicValidationInterval = 6 * time.Hour
+	maxPendingTaskAddSignals   = 4000
```
jaypipes: What's the reasoning behind this value?

wci/workflow/workflow.go
```diff
@@ -178,16 +184,24 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
 	}
 
-	// Process the signals from the prior run (pre-CaN)
-	d.processPendingTaskAddSignals(ctx)
+	useQueuedTaskAddSignals := workflow.GetVersion(ctx, taskAddSignalQueueProcessingPatch, workflow.DefaultVersion, 1) > workflow.DefaultVersion
+	d.limitPendingTaskAddSignals = workflow.GetVersion(ctx, taskAddSignalQueueLimitPatch, workflow.DefaultVersion, 1) > workflow.DefaultVersion
+	if !useQueuedTaskAddSignals {
+		// Process the signals from the prior run (pre-CaN)
+		d.processPendingTaskAddSignals(ctx)
+	}
```
jaypipes, line 206: Instead of all this new signal-queueing and entry-dropping behaviour, isn't there a way to add a single conditional check above `d.processPrendingTaskAddSignals()` to see if the history limit is approaching, and if so, just return the CaN error? This just seems quite complicated to implement such a simple conditional check.

wci/workflow/workflow.go
02strich: 500ms * 4000 signals = 30min'ish which seems like a reasonable latency limit while being conservative (in terms of not loosing signals). Not much more math/thought behind it. I'll add a comment explaining it as well.

wci/workflow/workflow.go
02strich: It was an incremental implementation artifact that there are two version markers here - reducing this to a single one for the new behavior.

What you are describing sounds like the code we had before this PR, or am I misunderstanding? The problem we have right now is that the check is not evaluated because the Select evaluates more than one signal so the condition isn't checked again (due to the concurrent signal handlers running). The change shortens the signal handler to just accepting the work and moves the actual work processing into a simple list processor that really just processes a single item at a time.

jaypipes: > What you are describing sounds like the code we had before this PR, or am I misunderstanding?

I don't believe so. The previous code always went into `processPendingTaskAddSignals()`, regardless of whether the history limit was approaching. I guess what I'm asking is why we couldn't just do this instead here:

```go
if d.historyLimitApproaching(ctx) {
    // Just CaN the workflow since we don't want to get near OOMing ourselves
	return temporal.NewApplicationError(iface.ErrLongHistory, iface.ErrLongHistory)
}

// Process the signals from the prior run (pre-CaN)
d.processPendingTaskAddSignals(ctx)
```

02strich: I am not following how this would help.

Before https://github.com/temporalio/temporal-auto-scaled-workers/blob/main/wci/workflow/workflow.go#L194 would only process signals that got handed over across a CaN boundary, which works fine and is unrelated to the issue at hand. Adding more conditions before that to trigger another CaN seems not useful (as we just CaN'd at that point). Before this PR the only thing that wrote pending signals was https://github.com/temporalio/temporal-auto-scaled-workers/blob/main/wci/workflow/workflow.go#L267

wci/workflow/workflow.go
```diff
@@ -760,6 +791,59 @@ func (d *WorkflowRunner) processPendingTaskAddSignals(ctx workflow.Context) {
...
+			"queue_size", len(d.State.PendingTaskAddSignals),
+			"max_queue_size", maxPendingTaskAddSignals,
+			"task_queue", req.TaskQueueName,
+			"task_queue_type", req.TaskQueueType.String(),
+			"sync_match", req.IsSyncMatch,
+			"sync_match_batch", req.SyncMatchSignalsSinceLast,
+			"no_sync_match_batch", req.NoSyncMatchSignalsSinceLast,
+		)
+		d.signalMetric(wcimetrics.SignalTypeTaskAdd).recordSkipped(wcimetrics.SkippedReasonQueueFull)
+		return
```
gcristea-temporal, line 815: This ignores recent/current task-add signals. Would it make sense to drop older entries from `d.State.PendingTaskAddSignals`?

gcristea-temporal (approved):
Approved, with one comment.

wci/workflow/workflow.go
02strich: I am not sure it makes a difference - both have risks. I decided to keep the old ones because they are first in the queue.

wci/workflow/workflow.go
seanbollin: I think @jaypipes is just wondering if the purpose of this PR is to stop it from running a history-limit-inducing amount of activities, can't there just be a check to see if we're close to that limit and if not do the same call to those activities as before?

seanbollin:
Were you able to load test this - meaning trigger the condition and see it work?

wci/workflow/workflow.go
02strich: The problem today is that you can't check that as the loop runs inside the `Select` API. The biggest part of the API is moving the loop out of there and into workflow code.

02strich:
> Were you able to load test this - meaning trigger the condition and see it work?

Yeah, @gcristea-temporal tested it

gcristea-temporal:
> > Were you able to load test this - meaning trigger the condition and see it work?
> 
> Yeah, @gcristea-temporal tested it

Only tested until I couldn't break it.

wci/workflow/workflow.go
```diff
@@ -190,16 +199,23 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 		return err
 	}
 
-	// Process the signals from the prior run (pre-CaN)
-	d.processPendingTaskAddSignals(ctx)
+	d.limitPendingTaskAddSignals = workflow.GetVersion(ctx, taskAddSignalQueueLimitPatch, workflow.DefaultVersion, 1) > workflow.DefaultVersion
+	if !d.limitPendingTaskAddSignals {
+		// Process the signals from the prior run (pre-CaN)
+		d.processPendingTaskAddSignals(ctx)
```
mani-j9, line 205: More of a theoretical question - I see how this kind of bounded approach helps in doing some meaningful work(processing task add signals) instead of CaN-ing again right after a CaN, but I wonder if there are any drawbacks to the approach of us trying to do work when it is time to CaN because we are not operating in an ideal situation(supposed to CaN).

02strich: I am not following the question. Can you explain the scenario a bit more?

In general, the line you highlighted is just processing a single signal that has been carried over b/c it was time to CaN. This is the implementation in production today.

mani-j9: Scenario I meant was when we need to CaN(as suggested) but we are processing the signals from the prior run(pre-CaN), technically that is not a feasible scenario because it takes sometime for the new instance to hit the history size limit. Not a valid concern, pls ignore

PR #107, opened 2026-08-07
Split error type on compute provider failures

zainawaisn:
## What was changed

`error_type` on `worker_controller_instance_activities` now distinguishes four kinds of compute provider failure instead of reporting everything as `compute_provider_failed`:

`compute_provider_misconfigured`: AWS rejected the customer's config (bad ARN, missing IAM permissions).
`compute_provider_service_unavailable`: AWS was down or unreachable.
`compute_provider_throttled`: Rate/concurrency limited. Config is valid, but capacity isn't available.
`internal`: WCI's own credentials or intermediary role chain.

compute_provider_failed remains as the fallback for anything unclassified.

Changelist:

- computeprovider.FailureClass is a small provider-agnostic set of judgments; ProviderError carries one alongside the original error. It's transparent for Error()/Unwrap(), so messages and error inspection are unchanged.
- classifyAWSFailure (aws.go) picks a class using the AWS SDK's own client/server fault tagging, its published throttle-code list, and its connection-error checks.
- buildBaseAWSConfig marks its four failure paths with errWCIOwned, since every input there is ours rather than the customer's.
- Lambda's InvokeWorker and ECS's UpdateWorkerSetSize wrap their results with a class. Bodies are untouched.
- computeProviderErrorType (activities.go) maps class → tag at both call sites.

Scope: Lambda and ECS. GCP Cloud Run will need its own classifier.

## Why?

One label was covering three (very) different failure situations for aws lambda invokes, so the metric could not solve the paging issue. We desire different alarms with different thesholds regarding customer errors, API downages, and WCI faults. ECS may require this in the future, so preemptively gave it this functionality.

Addressed PR feedback:
`misconfigured` → `rejected` + condition split, and `compute_provider_unavailable` → `compute_provider_not_enabled`. 

Also renamed `internal` → `compute_provider_internal`.

### DIFF
--- a/wci/metrics/metric_defs.go
+++ b/wci/metrics/metric_defs.go
@@ -90,28 +90,35 @@ const (
 	ErrorTypeInvalidSpec                           ErrorType = "invalid_spec"
 	// ErrorTypeActivityError is the coarse bucket used at workflow-side activity-execution failure sites, paired with
 	// the activity_error_type tag for sub-classification.
-	ErrorTypeActivityError              ErrorType = "activity_error"
-	ErrorTypeInvalidRequest             ErrorType = "invalid_request"
-	ErrorTypeAlgorithmUnavailable       ErrorType = "algorithm_unavailable"
-	ErrorTypeAlgorithmFailed            ErrorType = "algorithm_failed"
-	ErrorTypeComputeProviderUnavailable ErrorType = "compute_provider_unavailable"
+	ErrorTypeActivityError        ErrorType = "activity_error"
+	ErrorTypeInvalidRequest       ErrorType = "invalid_request"
+	ErrorTypeAlgorithmUnavailable ErrorType = "algorithm_unavailable"
+	ErrorTypeAlgorithmFailed      ErrorType = "algorithm_failed"
+	// ErrorTypeComputeProviderNotEnabled means no compute provider is enabled for the
+	// namespace, or the one the spec names is not registered. Nothing was attempted.
+	ErrorTypeComputeProviderNotEnabled ErrorType = "compute_provider_not_enabled"
 	// ErrorTypeComputeProviderFailed is the fallback for a compute provider failure
 	// the provider did not classify. The values below are the classified subsets.
 	ErrorTypeComputeProviderFailed ErrorType = "compute_provider_failed"
-	// ErrorTypeComputeProviderMisconfigured means the provider rejected the
-	// customer's config: a resource that does not exist, or no permission to it.
-	ErrorTypeComputeProviderMisconfigured ErrorType = "compute_provider_misconfigured"
+	// ErrorTypeComputeProviderRejected means the provider rejected a request made
+	// against the customer's config for a reason we could not narrow further.
+	ErrorTypeComputeProviderRejected ErrorType = "compute_provider_rejected"
+	// ErrorTypeComputeProviderRejectedNotFound means the compute resource named by
+	// the customer's config does not exist.
+	ErrorTypeComputeProviderRejectedNotFound ErrorType = "compute_provider_rejected_not_found"
+	// ErrorTypeComputeProviderRejectedAccessDenied means the resource exists but the
+	// role we assumed lacks permission to act on it.
+	ErrorTypeComputeProviderRejectedAccessDenied ErrorType = "compute_provider_rejected_access_denied"
 	// ErrorTypeComputeProviderServiceUnavailable means the provider's API was
-	// unreachable or returned a server-side error. Distinct from
-	// ErrorTypeComputeProviderUnavailable, which means the provider is not enabled
-	// for the namespace or not registered.
+	// unreachable or returned a server-side error.
 	ErrorTypeComputeProviderServiceUnavailable ErrorType = "compute_provider_service_unavailable"
 	// ErrorTypeComputeProviderThrottled means the provider rate- or
 	// concurrency-limited the request.
 	ErrorTypeComputeProviderThrottled ErrorType = "compute_provider_throttled"
-	// ErrorTypeInternal means the failure is attributable to worker-controller-instances's own
-	// configuration or credentials rather than the customer's.
-	ErrorTypeInternal ErrorType = "internal"
+	// ErrorTypeComputeProviderInternal means the compute provider call failed on
+	// worker-controller-instances's own configuration or credentials rather than the
+	// customer's.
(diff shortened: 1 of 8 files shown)


wci/metrics/metric_defs.go
```diff
@@ -95,7 +95,23 @@ const (
 	ErrorTypeAlgorithmUnavailable       ErrorType = "algorithm_unavailable"
 	ErrorTypeAlgorithmFailed            ErrorType = "algorithm_failed"
 	ErrorTypeComputeProviderUnavailable ErrorType = "compute_provider_unavailable"
-	ErrorTypeComputeProviderFailed      ErrorType = "compute_provider_failed"
+	// ErrorTypeComputeProviderFailed is the fallback for a compute provider failure
+	// the provider did not classify. The values below are the classified subsets.
+	ErrorTypeComputeProviderFailed ErrorType = "compute_provider_failed"
+	// ErrorTypeComputeProviderMisconfigured means the provider rejected the
+	// customer's config: a resource that does not exist, or no permission to it.
+	ErrorTypeComputeProviderMisconfigured ErrorType = "compute_provider_misconfigured"
```
ericbriffa: misconfigured is a (potential) cause, not a symptom. i think your comment actually the clarity this is intended to represent `a resource that does not exist, or no permission to it.`, those are more meaningful, can we raise them instead? or even "rejected" is more symptomatic and less presumptive of misconfiguration being the cause.

wci/metrics/metric_defs.go
```diff
@@ -95,7 +95,23 @@ const (
...
+	// ErrorTypeComputeProviderServiceUnavailable means the provider's API was
+	// unreachable or returned a server-side error. Distinct from
+	// ErrorTypeComputeProviderUnavailable, which means the provider is not enabled
```
ericbriffa: this would catch me out - would changing this type to "not enabled" be clearer?

wci/metrics/metric_defs.go
02strich: yeah, I would split this out into the actual conditions

wci/metrics/metric_defs.go
02strich: also seems like the comment needs to move to the right line....

wci/metrics/metric_defs.go
zainawaisn: renamed to `compute_provider_rejected` and split out `_rejected_not_found` / `_rejected_access_denied`. unrecognized codes stay in plain `rejected`.

wci/metrics/metric_defs.go
zainawaisn: renamed to compute_provider_not_enabled.

wci/workflow/compute_provider/aws.go
```diff
@@ -20,11 +21,89 @@ type stsAPI interface {
...
+		return FailureUnclassified
+	}
+
+	// A cancelled request tells us nothing about either axis; don't blame it on
+	// whoever's config happened to be in play.
+	if errors.Is(err, context.Canceled) {
+		return FailureUnclassified
+	}
+
+	// Ownership only separates the two client-fault buckets. A throttled or
```
jaypipes: Helpful hint when crafting code comments (or a hint to Claude if Claude wrote these)... if you are overly specific saying things like "the two client-fault buckets", the code comment becomes fragile in the face of the inevitable addition of addition client-fault buckets. :) Better to just say "Ownership separates client-fault from WCI-fault errors" instead of introducing almost immediately-stale comments...

wci/workflow/compute_provider/aws.go
```diff
@@ -39,15 +118,15 @@ func buildBaseAWSConfig(ctx context.Context, region string, intermediaryRoles []
...
 		}
 		if req.RoleSessionName == "" {
-			return aws.Config{}, fmt.Errorf("empty role session name for intermediary role")
+			return aws.Config{}, fmt.Errorf("%w: empty role session name for intermediary role", errWCIOwned)
 		}
 
 		awsConfig, err = assumeRoleWithRequest(ctx, awsConfig, &req)
 		if err != nil {
-			return aws.Config{}, err
+			return aws.Config{}, fmt.Errorf("%w: %w", errWCIOwned, err)
```
jaypipes:
```suggestion
			return aws.Config{}, fmt.Errorf("%w: error assuming intermediary role: %w", errWCIOwned, err)
```

jaypipes (approved):
I have received a PagerDuty alarm every 3 hours since Friday noon EST with this exact false alarm. I've slept in the guest bedroom the last 3 nights to keep from waking my wife on the predictable midnight, 3am, and 6am High Urgency phone calls from PagerDuty. It's the same customer(s) that have misconfigured their Lambda or Cloud Run compute configs and are throwing 404 Not Found errors which end up being logged as WCI scaling event failures.

From a purely selfish perspective, it would be great to get this merged ASAP. :)

wci/workflow/compute_provider/aws.go
zainawaisn: Done.

jaypipes (approved):
👍 thank you!

wci/workflow/compute_provider/aws_errors_test.go
```diff
@@ -0,0 +1,168 @@
...
+		// classifies as throttled.
+		{"too many requests", &types.TooManyRequestsException{}, FailureThrottled},
+		{"ec2 throttled", &types.EC2ThrottledException{}, FailureThrottled},
+		{"wci-owned throttle", wciOwned(&types.TooManyRequestsException{}), FailureThrottled},
+
+		// WCI-owned client faults are not narrowed: our own missing resource and our
+		// own denied permission are the same page for the same on-call.
+		{"wci-owned access denied", wciOwned(&mockFaultError{code: "AccessDenied", fault: smithy.FaultClient}), FailureInternal},
+		{"wci-owned not found", wciOwned(&types.ResourceNotFoundException{}), FailureInternal},
+		{"wci-owned local validation", wciOwned(errors.New("empty role session name")), FailureInternal},
```
ericbriffa, line 70: Why are we bundling all these different errors into one type? isn't this very similar to the problem we are trying to solve of having difficulty setting good alarm engagements because errors can be many different things?

jaypipes: @ericbriffa It's not bundling them into one type. It's simply creating different error codes/types that are categorized as originating from a problem in WCI as opposed to a user's misconfigured compute config.

PR #110, opened 2026-08-18
COM-214: Add release tagging

gcristea-temporal:
## What was changed

Adds release tags to the module. Tags are `v<major>.<minor>.<patch>-<serverVersion>`: WCI's semver core suffixed with the `go.temporal.io/server` build tag `go.mod` pins, flattened to digits and dots (server `v1.32.0-158.3` gives WCI `v1.0.0-1.32.0.158.3`). The suffix means a WCI version names the server build it was compiled against, so compatibility is readable from the version alone.

The tag is the whole release. Nothing publishes a GitHub Release — WCI is consumed as a Go library, so there are no artifacts to attach to one.

* Cutting a `release/vMAJOR.MINOR` branch is the only manual step. Its push runs `release-auto-tag.yml`: check the server pin, run the gates, compute the next patch, push the tag.
* The gates are the workflows PRs already run. `ci.yaml` and `golangci-lint.yaml` gained a `workflow_call` trigger and an optional `ref` input, so the release path calls them as `needs` dependencies instead of racing a parallel run. An empty `ref` means "the triggering ref", so `pull_request` and `push` are unaffected. That also puts `check-go-mod-sync` on the release path, which matters because a release requires the server pin to agree between `go.mod` and `tests/go.mod`.
* `.github/scripts/compute-release-tag.sh` assembles the tag: MAJOR.MINOR off the branch name, the highest existing patch on that line by version sort plus one, then the suffix. If HEAD already carries a tag on the line it returns that tag instead of a successor, so a job re-run at the same commit can't leave one commit with two release tags.
* `.github/scripts/resolve-server-suffix.sh` derives the suffix from the `go.mod` pin, falling back to the earliest server release containing the pinned commit when the pin is a pseudo-version.
* `docs/release.md` documents the process, the version scheme, and why the server version is flattened.
* Pins `actionlint` and `shellcheck` in `mise.toml` (`actionlint` only shell-lints `run:` blocks when `shellcheck` is on `PATH`).

## Why?

`temporal-auto-scaled-workers` has no tags, so its pseudo-version has a bare `v0.0.0-` base and `@latest` can only ever resolve to the tip of the default branch — there is no way to ask for a release. Improving the build process for Temporal Cloud requires a proper semver for the module.

## Changed since the last review

`release-tag.yml` and its GitHub Release publishing are gone, along with the `check-tag` job and the `workflow_dispatch` trigger on `release-auto-tag.yml`. WCI is a library and there are no artifacts to attach yet, so the tag is the release. That also removes the `push: tags` entry point, which could only ever withhold a Release: by the time it ran the tag already existed and was resolvable by `go get`.

Also in response to review: the inline tag-computation script moved out of the workflow into `.github/scripts/`, and re-running the tagger at an already-tagged commit now reuses the tag rather than minting a successor.

## Checklist

1. Closes 

3. How was this tested:

   Locally: `actionlint` and `shellcheck` clean; `compute-release-tag.sh` exercised against throwaway repos covering patch seeding, increment, version-sort ordering (`.10` over `.9`), reuse at an already-tagged commit, out-of-order tags on older commits, and rejection of malformed or suffix-less tags.

   The end-to-end path can't run before merge — the tagger only triggers on a `release/v*` push, and Actions are disabled for forks. Two throwaway-workflow dry runs validated the auto-tag mechanics, though they predate the changes above and so cover the earlier shape of the release path:
   - https://github.com/temporalio/temporal-auto-scaled-workers/actions/runs/32181923932
   - https://github.com/temporalio/temporal-auto-scaled-workers/actions/runs/32183412144

4. Any docs updates needed?

   Included: `docs/release.md`.

### DIFF
--- a/.github/scripts/resolve-server-suffix.sh
+++ b/.github/scripts/resolve-server-suffix.sh
@@ -1,0 +1,132 @@
+#!/usr/bin/env bash
+#
+# Resolve the server-release suffix for a WCI module tag.
+#
+# The WCI tag is v<wciMajor>.<wciMinor>.<wciPatch>-<serverVersion>, where
+# <serverVersion> is the pinned go.temporal.io/server build tag flattened to digits
+# and dots (v1.32.0-158.3 -> 1.32.0.158.3). This prints that suffix.
+#
+# Resolution:
+#   1. Read the go.temporal.io/server pin from go.mod (or arg $1).
+#   2. If it's already a clean build tag (vX.Y.Z-BUILD.REV), reformat and use it.
+#   3. Otherwise it's a pseudo-version (safety net): extract the commit hash, find
+#      the earliest server build tag that *contains* that commit, and use that. (Go
+#      bases a pseudo-version on the nearest ancestor tag, which can sit many builds
+#      behind — and on a different minor line — from where the commit really shipped.)
+#
+# Failure modes (both exit non-zero):
+#   - commit not reachable in temporal    -> can't build against it anyway
+#   - commit ahead of every release tag   -> WCI must track a *released* server
+#
+# Env:
+#   EMIT           "suffix" (default) prints the flattened suffix (1.32.0.158.3);
+#                  "tag" prints the resolved server build tag (v1.32.0-158.3) — used
+#                  by the release guard to suggest a repin target.
+#   TEMPORAL_DIR   use an existing temporal checkout instead of cloning (tests)
+#   GITHUB_TOKEN   auth for the clone (public repo; avoids the unauth rate limit)
+#   SERVER_REPO    override clone URL (default github.com/temporalio/temporal)
+set -euo pipefail
+
+# A semver numeric identifier: no leading zeros. Server build tag v1.32.0-158.03 would
+# flatten to the suffix 1.32.0.158.03, and `1.0.0-1.32.0.158.03` is not valid semver —
+# Go and the module proxy reject it. Fail here instead of at tag-push time.
+NUM='(0|[1-9][0-9]*)'
+RELEASE_TAG_RE="^v${NUM}[.]${NUM}[.]${NUM}-${NUM}[.]${NUM}$"
+
+# Flatten a server build tag to the WCI-tag suffix: v1.32.0-158.3 -> 1.32.0.158.3
+# (drop the leading v, turn the base/build separator '-' into '.').
+to_suffix() {
+  local tag="${1#v}"
+  printf '%s\n' "${tag/-/.}"
+}
+
+# Emit the resolved server build tag ($1): the flattened suffix by default, or the
+# tag itself when EMIT=tag (the guard uses that to suggest a repin target).
+emit() {
+  if [[ "${EMIT:-suffix}" == "tag" ]]; then
+    printf '%s\n' "$1"
+  else
+    to_suffix "$1"
+  fi
+}
+
+# Set only when we create a throwaway clone; removed on exit.
+CLONE_DIR=""
+cleanup() { [[ -n "$CLONE_DIR" ]] && rm -rf "$CLONE_DIR"; return 0; }
+trap cleanup EXIT
+
+server_pin() {
+  if [[ $# -ge 1 && -n "${1:-}" ]]; then
+    printf '%s\n' "$1"
+    return
+  fi
+  local ver
+  ver=$(grep 'go\.temporal\.io/server ' "${GOMOD:-go.mod}" | awk '{print $2}')
+  [[ -n "$ver" ]] || { echo "no go.temporal.io/server pin in ${GOMOD:-go.mod}" >&2; exit 1; }
+  printf '%s\n' "$ver"
+}
+
+# Earliest server release tag (by globally-monotonic build number) that contains
+# $1 in its history. Empty output if none.
+earliest_containing() {
+  local hash="$1"
+  git -C "$TEMPORAL_DIR" tag --contains "$hash" --list 'v*' 2>/dev/null \
+    | grep -E "$RELEASE_TAG_RE" \
+    | sort -t- -k2 -V \
+    | head -1
+}
+
+main() {
+  local ver
+  ver=$(server_pin "${1:-}")
+
+  # Already a clean build tag — no lookup needed.
+  if [[ "$ver" =~ $RELEASE_TAG_RE ]]; then
+    emit "$ver"
+    return
+  fi
+
+  # Build-tag shape but not canonical semver — say so, rather than reporting it as an
+  # unrecognised pin below.
+  if [[ "$ver" =~ ^v[0-9]+[.][0-9]+[.][0-9]+-[0-9]+[.][0-9]+$ ]]; then
+    echo "server pin '$ver' has a leading-zero component; the flattened suffix would not be valid semver" >&2
+    exit 1
+  fi
+
+  # Pseudo-version: the commit hash is the final dash-delimited segment.
+  local hash="${ver##*-}"
+  if [[ ! "$hash" =~ ^[0-9a-f]{12}$ ]]; then
+    echo "server pin '$ver' is neither a release tag nor a pseudo-version" >&2
+    exit 1
+  fi
+
+  # Get a commit graph (no trees/blobs) unless one was provided.
+  if [[ -z "${TEMPORAL_DIR:-}" ]]; then
+    TEMPORAL_DIR=$(mktemp -d)
+    CLONE_DIR="$TEMPORAL_DIR"
+    local url="${SERVER_REPO:-https://github.com/temporalio/temporal.git}"
+    if [[ -n "${GITHUB_TOKEN:-}" && "$url" == https://github.com/* ]]; then
+      url="https://x-access-token:${GITHUB_TOKEN}@${url#https://}"
+    fi
+    git clone --filter=tree:0 --no-checkout --quiet "$url" "$TEMPORAL_DIR"
+  fi
+
+  # Failure mode 1: commit not reachable in the repo.
+  if ! git -C "$TEMPORAL_DIR" cat-file -e "${hash}^{commit}" 2>/dev/null; then
+    echo "server commit ${hash} is not reachable in temporal (force-pushed, unmerged, or private)" >&2
+    exit 1
+  fi
+
+  local earliest
+  earliest=$(earliest_containing "$hash")
+
+  # Failure mode 2: reachable but not yet in any release.
+  if [[ -z "$earliest" ]]; then
+    echo "server commit ${hash} is not in any server release yet; WCI must track a released server build" >&2
+    exit 1
+  fi
+
+  emit "$earliest"
+}
+
+main "$@"
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,6 +1,13 @@
 name: Linting
 on:
   pull_request:
+  # Called by release-auto-tag.yml as a release gate, before it pushes a tag.
+  workflow_call:
+    inputs:
+      ref:
+        description: 'Commit, branch, or tag to lint (empty = the triggering ref)'
+        required: false
+        type: string
 
 permissions:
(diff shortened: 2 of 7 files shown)


docs/release.md
```diff
@@ -0,0 +1,99 @@
+# Releasing `temporal-auto-scaled-workers`
+
+A release is a git tag `v<major>.<minor>.<patch>-<serverVersion>` — WCI's own semver
+core plus the `go.temporal.io/server` build tag that `go.mod` pins, flattened to
+digits and dots (server `v1.32.0-158.3` → WCI `v1.0.0-1.32.0.158.3`). Downstream
```
02strich: can you add a note covering the reason for the flattening?

.github/scripts/resolve-server-suffix.sh
```diff
@@ -0,0 +1,132 @@
...
+  # Already a clean build tag — no lookup needed.
+  if [[ "$ver" =~ $RELEASE_TAG_RE ]]; then
+    emit "$ver"
+    return
+  fi
+
+  # Build-tag shape but not canonical semver — say so, rather than reporting it as an
+  # unrecognised pin below.
+  if [[ "$ver" =~ ^v[0-9]+[.][0-9]+[.][0-9]+-[0-9]+[.][0-9]+$ ]]; then
+    echo "server pin '$ver' has a leading-zero component; the flattened suffix would not be valid semver" >&2
```
02strich, line 92: what do you mean with 'leading-zero' here?

.github/workflows/release-auto-tag.yml
```diff
@@ -0,0 +1,186 @@
...
+      - name: Compute and push next tag
+        id: compute
+        env:
+          BRANCH: ${{ github.ref_name }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
```
02strich, line 114: out of curiosity: what's the default value for these in GHA? or is there none?

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,130 @@
...
+        run: |
+          set -euo pipefail
+          if gh release view "${TAG}" >/dev/null 2>&1; then
+            echo "Release ${TAG} already exists, skipping"
+            exit 0
+          fi
+          gh release create "${TAG}" \
+            --title "${TAG}" \
+            --generate-notes \
+            --verify-tag
```
02strich: any thoughts on immutable releases?

.github/scripts/resolve-server-suffix.sh
gcristea-temporal: A version containing a leading zero looks like `v1.0.02-1.32.0.158.3` (please note the leading zero in the PATCH.) That gets rejected by the semver parser
```
semver parse v1.0.02-1.32.0.158.3
error: Cannot parse version: v1.0.02-1.32.0.158.3
```

.github/workflows/release-tag.yml
gcristea-temporal: I think we should support that. However, adding them now would increase the complexity/effort. For now we are treating WCI as a library and create only tags. An immutable release should produce artifacts that can be used by customers.

.github/scripts/resolve-server-suffix.sh
02strich: `v1.0.02-1.32` is valid according the regexp in line 91, or am I misunderstanding something?

.github/workflows/release-auto-tag.yml
gcristea-temporal: They're not set (and that can lead to `git` using the default user, which is most likely to be the `<account>@<host>`)

.github/scripts/resolve-server-suffix.sh
gcristea-temporal: It does, and that is the intended behavior: if it a semver (line 84), the script exits. 
Line 91 filters out strings that look like a semver but they're really not
```
./scripts/resolve-server-suffix.sh v1.0.02-1.32
server pin 'v1.0.02-1.32' has a leading-zero component; the flattened suffix would not be valid semver
``` 
If it's not matching on line 91, it continues to trying to find a server release tag that has the input in its history. 

It actually tries to find the earliest such release tag, which may not be the best approach.

02strich: ah okay, that was not clear to me as logic flow

02strich (approved):

.github/workflows/golangci-lint.yaml
```diff
@@ -20,4 +30,6 @@ jobs:
         uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
           version: v2.12
-          args: --new-from-rev=${{ github.event.pull_request.base.sha }}
+          # PRs lint only what they changed; a release gate lints the whole tree,
+          # and base.sha does not exist off a pull_request event anyway.
```
jaypipes: If "base.sha does not exist off a pull_request event anyway", why is the conditional below referencing `github.event.pull_request.base.sha`?

.github/workflows/release-auto-tag.yml
```diff
@@ -0,0 +1,186 @@
...
+#   - The server release is a `-` prerelease suffix, not `+` build metadata: Go
+#     rejects build metadata for module versions (only +incompatible is legal), and
+#     downstream release-dependency checks allow only a -[0-9.]+ suffix.
+#   - Gates run *before* the tag push: a published tag is immediately resolvable by
+#     `go get` and the module proxy caches it immutably, so a tag that fails lint or
+#     tests cannot be retracted — only superseded.
+#   - The tag is pushed with GITHUB_TOKEN, so release-tag.yml is invoked via
+#     workflow_call rather than its push-tag trigger (a GITHUB_TOKEN tag push does
+#     not fire other workflows). It is told to skip the gates, which already ran
+#     here on this exact commit.
```
jaypipes: Good explanation/comments here, thank you! (Claude? :) )

.github/workflows/release-auto-tag.yml
```diff
@@ -0,0 +1,186 @@
...
+on:
+  push:
+    branches:
+      - 'release/v*'
+  workflow_dispatch:
```
jaypipes: Presumably we don't want any `input:` fields in this workflow because it's always triggered by the push of a release branch?

If that's the case, and there's no point having a manual override of the tag, I'd recommend removing the ability to manually execute this particular workflow.

.github/workflows/release-auto-tag.yml
```diff
@@ -0,0 +1,186 @@
...
+permissions:
+  contents: write # push the tag
+
+concurrency: # serialize so back-to-back pushes always tag off the previous tag
+  group: release-auto-tag-${{ github.ref_name }}
+  cancel-in-progress: false
+
+jobs:
+  # Seconds, and the most likely failure on a fresh release branch — so it runs
+  # before the multi-minute gates rather than alongside them.
```
jaypipes: > Seconds, and the most likely failure on a fresh release branch ...

The above seems like Claude interrupted its own train of thought :)

Perhaps it meant to say "This job runs before the multi-minute gate workflow jobs rather than alongside them since we expect this job to complete very quickly and, on failure, provide a short-circuit that will cancel the more expensive `ci` job."?

.github/workflows/release-auto-tag.yml
```diff
@@ -0,0 +1,186 @@
...
+          ref: ${{ github.sha }}
+          fetch-depth: 0
+
+      - name: Compute and push next tag
+        id: compute
+        env:
+          BRANCH: ${{ github.ref_name }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
```
jaypipes, line 111: Can you move this long Bash script to a separate `scripts/compute-release-tag.sh` file similar to the newly-added `scripts/resolve-server-suffix.sh` file?

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,130 @@
+name: Release
+
+# Validates a tagged commit and publishes a GitHub Release for it.
+#
+# Two entry points:
```
jaypipes: Technically there are three entry points :) `workflow_dispatch` is the third.

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,130 @@
...
+#   - workflow_call: invoked by release-auto-tag.yml right after it pushes a new
+#     v1.<serverBuild>.<wciRev> tag (a tag pushed with GITHUB_TOKEN does NOT fire
+#     the `push: tags` trigger below, so the auto-tag workflow chains here directly).
+#     That path gates *before* tagging and passes skip_gates.
+#   - push a `v*` tag by hand (with a human/PAT credential) → this runs directly, and
+#     runs the gates itself since nothing validated the commit first.
+#
+# Either way check-tag runs first: `push: tags` cannot be scoped to a branch, so the
+# "only release lines publish Releases" invariant is enforced as a job.
```
jaypipes: > "only release lines publish Releases" invariant is enforced as a job.

Not sure where the above "invariant" is documented. Was this in a Claude prompt maybe?

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,130 @@
...
+  contents: write # create the GitHub Release
+
+concurrency:
+  group: release-${{ inputs.tag || github.ref_name }}
+  cancel-in-progress: false
+
+jobs:
+  # Seconds, and both rejections are easy to trigger by hand — so they gate the
+  # runner time instead of surfacing after a full CI cycle. Mirrors check-pin in
+  # release-auto-tag.yml.
```
jaypipes: More Claude-speak :)

I don't know what "they gate in runner time" means. Maybe rewrite to a more human language:

> We quickly check that the tag being created is valid, allowing the more expensive `ci` job to be short-circuited if there is a problem.

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,130 @@
...
+          set -euo pipefail
+          # Numeric identifiers reject leading zeros — `v1.0.0-1.32.0.158.03` is not
+          # valid semver, so it could never resolve as a module version anyway.
+          NUM='(0|[1-9][0-9]*)'
+          if [[ ! "${TAG}" =~ ^v${NUM}[.]${NUM}[.]${NUM}-${NUM}([.]${NUM})+$ ]]; then
+            echo "Tag '${TAG}' is not a vMAJOR.MINOR.PATCH-<serverVersion> module tag" >&2
+            exit 1
+          fi
+
+          # `push: tags` cannot be scoped to a branch, so the invariant lives here:
```
jaypipes: LOL. Claude is obsessed with using the word "invariant".

docs/release.md
```diff
@@ -0,0 +1,99 @@
...
+git branch "${RELEASE_BRANCH}" "${BASE}"
+git push origin "${RELEASE_BRANCH}"    # ← triggers .github/workflows/release-auto-tag.yml
+```
+
+## What happens automatically after the push
+
+1. **`release-auto-tag.yml`** validates, then tags:
+   - `check-pin` guards that `go.mod`'s server pin is a build-tag release; on a
+     pseudo-version it fails and prints the exact repin command (via
+     `resolve-server-suffix.sh`). Seconds, and first, so a bad pin fails fast.
```
jaypipes: There's that "Seconds, and first, so a bad pin fails fast" line... please remove it or just write "Fails fast on an invalid Temporal server go.mod dependency."

.github/workflows/release-auto-tag.yml
gcristea-temporal: Moved to a script and moved all (two) scripts to `.github/scripts`

jaypipes (approved):
👍 thanks @gcristea-temporal :)

PR #111, opened 2026-08-18
Allow configuring subprocess provider timeout

Falven:
## Summary

- add an optional `timeout` setting to the development-only subprocess provider
- validate positive Go duration strings and pass the duration to GNU `timeout`
- preserve the existing one-minute default and termination behavior

## Why

The subprocess provider hardcodes a one-minute timeout. This interrupts development and system-test workers whose bounded work takes longer, and there is no way to adjust the limit per scaling group.

## User impact

Subprocess users can set `compute.config.timeout` to a value such as `30m` or `1h`. Existing configurations continue to use `1m`.

## Validation

- `go test ./wci/workflow/compute_provider`

### DIFF
--- a/wci/workflow/compute_provider/subprocess.go
+++ b/wci/workflow/compute_provider/subprocess.go
@@ -8,17 +8,19 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
+	"time"
 
 	"go.temporal.io/auto-scaled-workers/wci/workflow/iface"
 	"go.temporal.io/server/common/dynamicconfig"
 )
 
 const (
-	// subprocessTimeoutArg is the argument passed to the timeout(1) command (GNU coreutils).
-	subprocessTimeoutArg = "1m"
+	// defaultSubprocessTimeoutArg is the default argument passed to timeout(1) (GNU coreutils).
+	defaultSubprocessTimeoutArg = "1m"
 
 	configSubprocessCommand = "command"
 	configSubprocessArgs    = "args"
+	configSubprocessTimeout = "timeout"
 )
 
 type subprocessComputeProvider struct{}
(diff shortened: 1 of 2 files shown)


wci/workflow/compute_provider/subprocess.go
```diff
@@ -8,17 +8,19 @@ import (
...
+	"time"
 
 	"go.temporal.io/auto-scaled-workers/wci/workflow/iface"
 	"go.temporal.io/server/common/dynamicconfig"
 )
 
 const (
-	// subprocessTimeoutArg is the argument passed to the timeout(1) command (GNU coreutils).
+	// subprocessTimeoutArg is the default argument passed to timeout(1) (GNU coreutils).
 	subprocessTimeoutArg = "1m"
```
02strich:
```suggestion
	defaultSubprocessTimeoutArg = "1m"
```

02strich (commented):
Thank you for the contribution!

I have left a small comment, otherwise ready to get it in!

Falven:
> Thank you for the contribution!
> 
> I have left a small comment, otherwise ready to get it in!

@02strich Since issues are disabled here, could I ask a related design question? Is worker lifetime meant to stay provider-specific, or should WCI expose it per scaling group? We were considering a separate lifetime setting that gives each launch an absolute deadline. Supported providers would enforce and pass it to the worker so it can stop polling and drain. Does that fit WCI's direction, or should lifetime remain in the specific provider's config? I wanted to make this fix more comprehensive and provider-agnostic, but I think it would require changing both this repo and the main repo.

02strich (approved):

02strich:
@Falven thank you for the update. I have enabled issues now (thank you for calling that out!).

It's an interesting question on lifetime. Today the setting is provider specific, b/c we see that constrained lifetime is mostly a Lambda thing. Other providers (e.g. CloudRun, ECS or Kubernetes) do not have that constraint. What is the background of looking to have lifetime limits?

Outside of that, the client lifetime modeling is client-only at this point. This means that the lifetime that you set with WCI and Lambda need to be the same. For other providers, there is no provide-level timeout setting so it would be purely in the worker code. We are thinking about server provided settings towards workers, which would be a more generic solution to provide such settings to clients (and would be compute provider independent).

PR #115, opened 2026-08-20
Log task queue names alongside scaling metrics snapshot

johanngan:
## What was changed
Add the first 3 task queue names and task queue count to logging for scaling metrics snapshots.

## Why?
This should make it easier to associate worker deployment versions with the task queues they poll. In most cases we should expect relatively few task queues per WDV, but also log the task queue count so we can tell if there's omissions in logging.

## Checklist

1. How was this tested:
Ran workflow through local temporal server with serverless workers configured, viewed logs. Example log entry (some fields omitted)

```json
{"level":"info",msg":"Pulled scaling metrics snapshot","workflow_count":0,"activity_count":0,"nexus_count":0,"task_queue_count":2,"topk_task_queues":["[Workflow]test-tasks","[Activity]test-tasks"]}
```

### DIFF
--- a/wci/workflow/scaling_metrics_snapshot.go
+++ b/wci/workflow/scaling_metrics_snapshot.go
@@ -36,6 +36,9 @@ func (a *Activities) pullScalingMetricsSnapshot(ctx context.Context, namespaceNa
 		Activity: &iface.QueueTypeScalingMetrics{},
 		Nexus:    &iface.QueueTypeScalingMetrics{},
 	}
+	taskQueueCount := 0
+	const taskQueueSampleMax = 3
+	taskQueueSample := []string{}
 	for _, versionedTaskQueue := range deploymentVersionDetails.VersionTaskQueues {
 		if versionedTaskQueue == nil || versionedTaskQueue.Stats == nil {
 			continue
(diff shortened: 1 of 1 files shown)


wci/workflow/scaling_metrics_snapshot.go
```diff
@@ -61,9 +65,23 @@ func (a *Activities) pullScalingMetricsSnapshot(ctx context.Context, namespaceNa
 			// using unspecified instead of default to ensure (future) unhandled values are caught by linter.
 			logger.Warn("Unspecified task queue type for scaling metrics snapshot (namespace: %s, deployment: %s, buildId: %s).", namespaceName, deploymentName, deploymentBuildID)
```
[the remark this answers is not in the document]
johanngan: Unrelated to this change, but we should fix this is a different PR

wci/workflow/scaling_metrics_snapshot.go
```diff
@@ -11,6 +11,7 @@ import (
 	"go.temporal.io/auto-scaled-workers/wci/workflow/iface"
 	scalingalgorithm "go.temporal.io/auto-scaled-workers/wci/workflow/scaling_algorithm"
 	"go.temporal.io/sdk/activity"
+	"go.temporal.io/server/common/log/tag"
 )
```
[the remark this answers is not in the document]
johanngan: I wanted to avoid reflection, but it does work without the tagging 🤷. Removed it for simplicity.

johanngan:
Lint job is failing because Go 1.27.0 was released yesterday, and our job config is buggy, see https://github.com/temporalio/temporal-auto-scaled-workers/pull/116

wci/workflow/scaling_metrics_snapshot.go
```diff
@@ -61,9 +64,23 @@ func (a *Activities) pullScalingMetricsSnapshot(ctx context.Context, namespaceNa
 			// using unspecified instead of default to ensure (future) unhandled values are caught by linter.
 			logger.Warn("Unspecified task queue type for scaling metrics snapshot (namespace: %s, deployment: %s, buildId: %s).", namespaceName, deploymentName, deploymentBuildID)
 		}
+
+		taskQueueCount++
+		// Unclear what we should rank task queues by here; for now just use array order
```
jaypipes: Since the list of task queues returned in `deploymentVersionDetails.VersionTaskQueues` has no sort order (as you've implied here, and you are correct that there is [no indication in the API spec that there is any sort here](https://github.com/temporalio/api/blob/5520164fddb4c7c97592007dda82030fc4dcea19/temporal/api/workflowservice/v1/request_response.proto#L2495-L2496)), the key in the log value should not be `topk_task_queues`.

Unless, of course, you make a judgment call here and decide on a ranking method. My recommendation would be rank by backlog count since we already output backlog count sums in the log record. Then, change the log key to something like `task_queues_topk_backlog_count` and include the backlog count in the log value along with the task queue's type and name. Also make sure to sort the list by the dimension you choose to rank by. :)

jaypipes (changes requested):
Hi @johanngan, thanks for this! :) I've got a suggestion inline for ya.

wci/workflow/scaling_metrics_snapshot.go
johanngan: Yeah I didn't like the name either. Was debating something like `task_queue_sample` or `task_queue_subset` but I thought `topk` made the truncation aspect more obvious. Maybe just `task_queues` since we're logging the total count anyway?

I was deliberately trying to keep the name vague (on both the k value and what's being sorted by); that way if we later decide that it makes sense to rank by something else or change the k value, the log key won't have to change, which seems convenient for log search. Though perhaps the opposite thing would be better, where a change in ranking corresponded to a change in key. Any thoughts?

Good point about already logging the backlog counts, I suppose that distinguishes it from the other metrics somewhat. My thinking was that since in most cases I wouldn't expect to have more than 3 task queues anyway (at most 1 per type), sorting is probably unnecessary complexity. Feel free to correct me on that assumption though.

02strich: I think that log outputs backlog count is a bit of a historical bias more than it being a super relevant piece of information - it might be useful to expand on that as well at some point.

For the task queue names, I would be partial to `task_queue_sample` or similar to show the sampling aspect but not bother with any ordering as the meaningful ordering dimension is quite situation specific.

johanngan: Sounds good, changed the key to `task_queue_sample`

02strich (approved):

jaypipes (approved):
👍 I'm cool with `task_queue_sample` since there's no ranking or sort order.

PR #116, opened 2026-08-20
CI: golangci-lint: Use go.mod for Go version

johanngan:
## What was changed
Use go.mod for Go version in golangci-lint job instead of the latest stable.

## Why?
Prevents the version from moving under us when Go releases new versions, and ties it to our go.mod file like all the other CI jobs.


jaypipes (approved):
👍

PR #117, opened 2026-08-20
Introduce AgentCore compute provider

justinschoeff:
This adds a new compute provider that invokes an AWS Agent Core Runtime
endpoint in customers account using the same assume role mechanisms as
AWS Lambda provider.

## What was changed
This adds a new compute provider that invokes an AWS Agent Core Runtime
endpoint in customers account using the same assume role mechanisms as
AWS Lambda provider.

## Why?
To support AWS AgentCore Runtime as a compute provider to enable agentic workflows with Temporal on AWS. 

## Checklist

1. Closes 

2. How was this tested:

Local server invoked real AWS AgentCore Runtime.

3. Any docs updates needed?


wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -0,0 +1,273 @@
...
+	requireRoleAndExternalID bool
+}
+
+type agentCoreDataAPI interface {
+	InvokeAgentRuntime(ctx context.Context, params *agentcore_dp.InvokeAgentRuntimeInput, optFns ...func(*agentcore_dp.Options)) (*agentcore_dp.InvokeAgentRuntimeOutput, error)
+}
+
+type agentCoreControlAPI interface {
+	GetAgentRuntimeEndpoint(ctx context.Context, params *agentcore_cp.GetAgentRuntimeEndpointInput, optFns ...func(*agentcore_cp.Options)) (*agentcore_cp.GetAgentRuntimeEndpointOutput, error)
+}
```
jaypipes, line 46: good to see an API with clean separation of control and data plane :)

wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -0,0 +1,273 @@
...
+	// The endpoint name is passed through as the AgentCore invoke Qualifier.
+	input := &agentcore_dp.InvokeAgentRuntimeInput{
+		AgentRuntimeArn: aws.String(params.RuntimeARN),
+		ContentType:     aws.String("application/json"),
+		Payload:         payload,
+		Qualifier:       aws.String(params.RuntimeEndpoint),
+	}
+
+	// AgentCore doesn't have an async invocation method. This aims to make the call and close the response stream on
+	// our end as soon as possible, preventing this activity from blocking.
```
jaypipes, line 142: Hmm, interesting. So, the API call is `InvokeAgentRuntime` but it doesn't operate like Lambda's `InvokeFunction` works. That is... annoyingly inconsistent but par for the course for AWS services :)

wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -0,0 +1,273 @@
...
+	eid, _ := cfg[configAWSAgentCoreRoleExternalID].(string)
+	if roleARN == "" || eid == "" {
+		return nil
+	}
+	region, err := extractRegionFromARN(runtimeARN)
+	if err != nil {
+		return fmt.Errorf("cannot verify external ID enforcement: failed to extract region from AgentCore runtime ARN %q: %w", runtimeARN, err)
+	}
+	return verifyExternalIDEnforcedFn(ctx, region, roleARN, p.intermediaryRoles)
+}
```
jaypipes, line 172: future cleanup... might be cool to move this to a function shared between all AWS compute providers.

jaypipes (approved):
Nice. No concerns from me. Looks very clean. Thanks @justinschoeff!

wci/workflow/compute_provider/aws_agentcore.go
justinschoeff: Both `extractRegionFromARN` and `verifyExternalIDEnforcedFn` are coming from aws.go and shared with lambda/ecs. I guess we could share the cfg keys as they are the same values,b ut apart from parsing the cfg, this is mostly shared.

wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -0,0 +1,273 @@
...
+	return NewProviderError(classifyAWSFailure(err), err)
+}
+
+func (p *awsAgentCoreComputeProvider) invokeWorker(ctx context.Context, rc RequestContext, cfg ComputeProviderConfig) error {
+	invokeClient, params, err := p.getDataClientAndParams(ctx, cfg)
+	if err != nil {
+		return err
+	}
+
+	// Provide worker deployment version details in the payload
```
02strich, line 121: Is there a specific reason that AgentCore needs this more than the others? I would bias for consistency otherwise.

wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -0,0 +1,273 @@
...
+	input := &agentcore_dp.InvokeAgentRuntimeInput{
+		AgentRuntimeArn: aws.String(params.RuntimeARN),
+		ContentType:     aws.String("application/json"),
+		Payload:         payload,
+		Qualifier:       aws.String(params.RuntimeEndpoint),
+	}
+
+	// AgentCore doesn't have an async invocation method. This aims to make the call and close the response stream on
+	// our end as soon as possible, preventing this activity from blocking.
+	resp, err := invokeClient.InvokeAgentRuntime(ctx, input)
```
02strich, line 143: I am a bit confused about the comment: this only starts the session and is async by nature, no?

wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -0,0 +1,273 @@
+package computeprovider
```
02strich, line 1: should this new provider be shipped to customers, or marked as testing for the moment?

justinschoeff: I’ll follow up with a tag for testing.

wci/workflow/compute_provider/aws_agentcore.go
justinschoeff: Sure. I don't feel strongly, but like having it. Am happy to revert to empty payload aligning with others.

wci/workflow/compute_provider/aws_agentcore.go
justinschoeff: No, InvokeAgentRuntime will return whatever the HTTP endpoint from your container responds with. Think of sending a user message to a chat agent in agentcore which calls llm, performs some tools, LLM again and then streams response back to caller, this would contain that stream.

The way we are builder workers, this response won't be valuable apart from status code, but customers could do all sorts of nonsense and block here and we don't want to wait for it.

wci/workflow/compute_provider/aws_agentcore.go
justinschoeff: Following up here. I'm planning on testing in cell for docs / UI guys to work with, so I think we want it as is.

wci/workflow/compute_provider/aws_agentcore.go
02strich: Yeah, I think we should do it as well but not in a 1-off fashion so lets stick to empty here for now as well plz

wci/workflow/compute_provider/aws_agentcore.go
02strich: Ah that's the thought process - makes sense. I am not sure how much your approach really catches if the API is meant to be sync. We should make sure in our examples/framing that very little happens in the sync part of the invocation on the agent side.

PR #120, opened 2026-08-24
Remove standalone worker binary

02strich:
## What was changed
The worker binary and it's make targets have been removeed.

## Why?
It is neither used, nor working at this point so removing it.

If you are running WCI with open-source temporal, it is compiled into the temporal server binary and doesn’t use this worker.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


smuneebahmad (approved):

jaypipes (approved):
well that was quick :)

johanngan (approved):

PR #121, opened 2026-08-24
Add loggercheck to golangci-lint

johanngan:
## What was changed
Enable the loggercheck linter, which is built into golangci-lint.

## Why?
This allows us to catch common structured logging errors with the Temporal SDK logger.

## Checklist
1. How was this tested:
Tested against the bad log call from https://github.com/temporalio/temporal-auto-scaled-workers/pull/119

```
wci/workflow/scaling_metrics_snapshot.go:65:122: odd number of arguments passed as key-value pairs for logging (loggercheck)
                        logger.Warn("Unspecified task queue type for scaling metrics snapshot (namespace: %s, deployment: %s, buildId: %s).", namespaceName, deploymentName, deploymentBuildID)
```

```
wci/workflow/scaling_metrics_snapshot.go:65:16: logging message should not use format specifier "%s" (loggercheck)
                        logger.Warn("Unspecified task queue type for scaling metrics snapshot (namespace: %s, deployment: %s).", namespaceName, deploymentName)
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,6 +20,7 @@ linters:
     # - gomodguard
     - govet
     - iotamixing
+    - loggercheck
     - staticcheck
     - testifylint
     - usetesting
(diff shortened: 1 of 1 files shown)


02strich (approved):

jaypipes (approved):
thank you, thank you. :)

PR #126, opened 2026-09-03
Add scale suppression when dispatch rate is within epsilon detection to no-sync scaling algo

mani-j9:
## What was changed

Adds queue metrics based flat dispatch rate detection to no-sync scaling algorithm, gated on `scale_up_dispatch_rate_epsilon > 0`:

- A metrics poll confirms the queue's dispatch rate staying flat under a material backlog and persists a suppression lease.
- The lease gates the scale-up on **both** the poll path and the task-add fast path. 
- `epsilon = 0` (default) is unchanged from today's behavior.

## Why?

This is to suppress the scale up actions that are based on the presence of no-sync matches from matching outcome or backlog whenever we observe that the dispatch rate is trending flat even with workers being added. 

## Checklist

1. Closes COM-366
2. How was this tested:

Ran a load that drives arrival > dispatch rate and simulates the flat dispatch rate behavior, validated that suppression engages. 

3. Any docs updates needed?

Yes, will follow up with the doc and CLI changes in the next PRs

### DIFF
--- a/docs/simulators/function/SPEC.md
+++ b/docs/simulators/function/SPEC.md
@@ -10,8 +10,9 @@ existing workers are not keeping up. It has two scale-up paths:
 
 - Task-add signals from Matching, including direct no-sync matches and batched
   no-sync matches.
-- Periodic task-queue metric polls, based on backlog, worker lifetime refresh,
-  and dispatch-rate stability.
+- Periodic task-queue metric polls, based on backlog and worker lifetime
+  refresh, with an optional per-queue flat-dispatch-rate detector that gates
+  scale-up at a downstream/rate-limit ceiling.
 
 The algorithm does not scale down and does not set a target worker-set size.
 
@@ -24,14 +25,26 @@ The algorithm accepts only these config keys. Unknown keys are invalid.
 | `scale_up_cooloff_ms` | integer | `100` | `>= 0` | Minimum elapsed time between scale-up actions on the backlog-threshold path and task-add path. `0` disables cooloff. |
 | `scale_up_backlog_threshold` | integer | `0` | `>= 0` | Periodic metric polls request scale-up only when backlog is strictly greater than this threshold. |
 | `max_worker_lifetime_ms` | integer | `600000` | `>= 0` | Periodic metric polls request worker refresh when backlog is present and this much time has elapsed since the last scale-up. `0` disables this path. |
-| `scale_up_dispatch_rate_epsilon` | number | `0` | `>= 0` | Periodic metric polls suppress otherwise eligible scale-up when dispatch rate is unchanged within this epsilon. `0` disables suppression. |
+| `scale_up_dispatch_rate_epsilon` | number | `0` | `>= 0`, and `<= 0.10` when `> 0` | Enables the flat-dispatch-rate detector. `0` disables it. When `> 0` it is a **relative band** (a fraction of the dispatch rate): a queue's dispatch rate must stay within `epsilon * reference_rate` to count as flat. |
 | `metrics_poll_interval_ms` | integer | `60000` | `>= 10000` | Requested interval before the next periodic metric poll. |
+| `flat_dispatch_rate_confirm_ms` | integer | `90000` | `> metrics_poll_interval_ms` when `epsilon > 0` | A queue's dispatch rate must stay flat under a material backlog at least this long before suppression engages. Must exceed one poll interval, else suppression fires on the second flat poll regardless and the knob has no effect. |
+| `suppress_scale_up_ms` | integer | `120000` | `> suppress_poll_interval_ms` when `epsilon > 0` | Lifetime (TTL) of the suppression lease. Each flat poll renews it; if polling stops it self-expires after this. |
+| `suppress_poll_interval_ms` | integer | `90000` | `> 0` when `epsilon > 0` | Poll cadence while actively suppressing. Deliberately longer than `metrics_poll_interval_ms`. |
 
 Cross-field validation:
 
 - If `scale_up_cooloff_ms > 0`, `metrics_poll_interval_ms` must be greater than
   or equal to `scale_up_cooloff_ms`.
 - If `scale_up_cooloff_ms == 0`, the cross-field check is skipped.
+
+When `scale_up_dispatch_rate_epsilon > 0`:
+
+- `scale_up_dispatch_rate_epsilon` must be `<= 0.10` (a relative fraction).
+- `flat_dispatch_rate_confirm_ms` must be `> metrics_poll_interval_ms` (at or
+  below one poll interval the confirm window has no effect).
+- `suppress_poll_interval_ms` must be `> 0`.
+- `suppress_scale_up_ms` must be `> suppress_poll_interval_ms`, so the lease
+  never lapses between polls.
 
 ## State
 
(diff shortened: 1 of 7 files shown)


jaypipes:
> * A metrics poll confirms the queue's dispatch rate staying flat under a material backlog and persists a suppression lease.

What does the above mean? What is "a material backlog"? What is "a suppression lease"? :/

docs/simulators/function/simulator.js
```diff
@@ -249,55 +277,74 @@
...
+    gatingStatus = suppressed ? 'suppressed' : (flatSinceMs !== 0 ? 'confirming' : 'clear');
+    return suppressed;
+  }
 
-    if (candidate) {
-      invokeWorker('Metrics poll: ' + reason);
-      return;
-    }
+  // One scale-up per poll: growth (gated by the suppression verdict) OR maintenance (lifetime refresh, never gated).
+  // Cadence backs off while suppressing. epsilon <= 0 -> detectFlatDispatch returns false -> exactly today's behavior.
```
jaypipes: "suppression verdict"?

docs/simulators/function/simulator.js
```diff
@@ -249,55 +277,74 @@
...
+    const now = Date.now();
+    const backlog = queue.length;
+    const rate = currentDispatchRate(now);
+
+    const material = backlog > config.scaleUpBacklogThreshold;
+    const band = config.scaleUpDispatchRateEpsilon * refRate;         // RELATIVE band (fraction of dispatch)
+    const moved = refRate >= 0 && Math.abs(rate - refRate) > band;    // dispatch rose OR dropped
+
+    if (!material || moved || rate <= 0) {
+      suppressUntilMs = 0; flatSinceMs = 0; refRate = -1;             // resume: clear the verdict
```
jaypipes: "clear the verdict"?

docs/simulators/function/simulator.js
```diff
@@ -249,55 +277,74 @@
...
-    } else if (backlog > config.scaleUpBacklogThreshold) {
-      reason = 'Scale-up cooloff';
-    }
-
-    if (!candidate && config.maxWorkerLifetimeMs > 0 && backlog > 0 && elapsed >= config.maxWorkerLifetimeMs) {
-      candidate = true;
-      reason = 'Worker lifetime refresh';
+  // Confirm whether dispatch is at a flat-rate ceiling; persists the verdict (flat_since, ref_rate, suppress lease)
+  // that the fast path reads, and returns whether growth is currently suppressed. epsilon <= 0 clears the verdict and
+  // returns false, reverting to today's behavior. Mirrors detectDispatchCeiling in the Go algorithm (no branch-off).
```
jaypipes: No idea what "reverting to today's behavior" means. And, Claude will not know what this means two months from now. :)

docs/simulators/function/SPEC.md
```diff
@@ -24,25 +25,41 @@ The algorithm accepts only these config keys. Unknown keys are invalid.
...
 | `max_worker_lifetime_ms` | integer | `600000` | `>= 0` | Periodic metric polls request worker refresh when backlog is present and this much time has elapsed since the last scale-up. `0` disables this path. |
-| `scale_up_dispatch_rate_epsilon` | number | `0` | `>= 0` | Periodic metric polls suppress otherwise eligible scale-up when dispatch rate is unchanged within this epsilon. `0` disables suppression. |
+| `scale_up_dispatch_rate_epsilon` | number | `0` | `>= 0`, and `<= 0.10` when `> 0` | Enables the flat-dispatch-rate detector. `0` disables it. When `> 0` it is a **relative band** (a fraction of the dispatch rate): a queue's dispatch rate must stay within `epsilon * reference_rate` to count as flat. |
 | `metrics_poll_interval_ms` | integer | `60000` | `>= 10000` | Requested interval before the next periodic metric poll. |
+| `flat_dispatch_rate_confirm_ms` | integer | `90000` | `> metrics_poll_interval_ms` when `epsilon > 0` | A queue's dispatch rate must stay flat under a material backlog at least this long before suppression engages. Must exceed one poll interval, else suppression fires on the second flat poll regardless and the knob has no effect. |
```
jaypipes: > A queue's dispatch rate must stay flat under a material backlog at least this long before suppression engages. Must exceed one poll interval, else suppression fires on the second flat poll regardless and the knob has no effect.

What is "flat under a material backlog"? "else suppression fires"? "before suppression engages"?

This is another case of AI using technical-sounding terms without actually defining what the term means. :(

Searching in this repository for the term "material backlog", I only see this in the `docs/simulator/workset-ewma/SPEC.md` file:

```
`Material backlog threshold`: backlog length at or above which cadence may sample per-worker capacity.
```

So I guess "material backlog" means "when a task queue whose backlog depth is greater than some configured value when the scaler simulator's cadence ticker fires"?

It uses the phrase "flat under a material backlog" but the term "flat" implies a constant rate. That is not what is being described here, actually. What is being described here is that the backlog depth should remain within the range calculated from looking at the `scale_up_dispatch_rate_epsilon` value and the actual value of the backlog depth.

I sure wish Claude would use more direct and clear wording. Something like this would be much easier to comprehend than the pseudo-scientific jargon it wrote above:

> `rate_limited_dispatch_window_ms` | integer | `90000` | `> metrics_poll_interval_ms` when `scale_up_dispatch_rate_epsilon > 0` | The amount of time the scaling algorithm uses to track the backlog size when using the rate-limited approach to scale up. Each window of time of this length, the scaling algorithm will check to see if the backlog depth for the target queue is within the range given by the product of the backlog depth and `scale_up_dispatch_rate_epsilon`. If the product is greater than `material_backlog_threshold`, the scaler backs off on scaling up until the next evaluation window.

jaypipes (changes requested):
I really dislike the whole "flat dispatch" wording here. There's nothing flat about it and the term disguises the underlying mechanism which is *rate limiting the call to scale up a serverless worker*.

@02strich @mani-j9 can we discuss the naming of these things since these will become part of the public API (at least the metrics part...)?

wci/workflow/scaling_algorithm/no_sync_match.go
```diff
@@ -60,15 +90,30 @@ var noSyncValidConfigKeys = map[string]struct{}{
...
+	configNoSyncSuppressScaleUpMsKey:          {},
+	configNoSyncSuppressPollIntervalMsKey:     {},
 }
 
-var noSyncValidStateKeys = map[string]struct{}{
-	stateLastScaleUpTimestampKey:                         {},
-	fmt.Sprintf(stateLastDispatchRateKeyFmt, "workflow"): {},
-	fmt.Sprintf(stateLastDispatchRateKeyFmt, "activity"): {},
-	fmt.Sprintf(stateLastDispatchRateKeyFmt, "nexus"):    {},
+// queueTypeName maps each task queue type to its per-queue state-key prefix; an unrecognized type maps to "".
```
seanbollin: These Claude generated comments are not so good. I put it in the AGENTS.md file to not do this - but I guess it isn't working?

To me it's obvious looking at the code it's a map of queue types - do we even need a comment here?

docs/simulators/function/simulator.js
seanbollin: Agree 100% I cannot understand what this is by reading it

seanbollin:
Is there any Notion doc about this feature?

mani-j9:
> I really dislike the whole "flat dispatch" wording here. There's nothing flat about it and the term disguises the underlying mechanism which is _rate limiting the call to scale up a serverless worker_.
> 
> @02strich @mani-j9 can we discuss the naming of these things since these will become part of the public API (at least the metrics part...)?

@jaypipes  Rate limited task queue dispatch is one scenario which can result in such flat(within an epsilon band) dispatch rate trend even when we add workers as our reaction to backlog > threshold. I am hesitant in naming it rate_limit related because we aren't enforcing any limit here, we are using a proxy indicator to observe a world where adding workers to burn backlog is not effective, to help us decide that adding more workers won't help. 

I went for flat dispatch rate to indicate that the dispatch rate isn't moving as per the workers we are adding, iow the movement is negligible within an epsilon band.

wci/workflow/scaling_algorithm/no_sync_match.go
mani-j9: agree, will remove it

docs/simulators/function/simulator.js
mani-j9: I missed this, my bad. will remove it. Today's behavior is referring to the default epsilon = 0 branch, but the today term is misleading.

docs/simulators/function/SPEC.md
mani-j9: >It uses the phrase "flat under a material backlog" but the term "flat" implies a constant rate. That is not what is being described here, actually. What is being described here is that the backlog depth should remain within the range calculated from looking at the scale_up_dispatch_rate_epsilon value and the actual value of the backlog depth.

The algorithm tracks the trend of dispatch rate while adding workers to help process that observed backlog, it determines if the scale up is ineffective using dispatch_rate as the indicator to evaluate the effectiveness of adding workers.

mani-j9: @jaypipes I renamed "flat dispatch" usage to just say "dispatch rate within epsilon" instead in the latest rev. I considered "ineffective scale up", seems too broad.

I've considered other options but most of them(ceiling, unchanged, etc) seem similar to flat so I figured I'd just keep it simple. 

let me know if you have any other naming suggestions. 

edit: i'm yet to address the test cases to follow the same naming, will do it.

PR #129, opened 2026-09-10
Change default compute provider config behavior

johanngan:
## What was changed
Default to an empty compute provider allowlist rather than a full one.

## Why?
This makes the config behave as written, instead of assigning a special meaning to the null value.

### DIFF
--- a/wci/workflow/compute_provider/subprocess.go
+++ b/wci/workflow/compute_provider/subprocess.go
@@ -1,4 +1,4 @@
-//go:build !release
+//go:build computeprovider_subprocess && !release
 
 package computeprovider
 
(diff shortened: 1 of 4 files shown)


wci/workflow/compute_provider/subprocess.go
```diff
@@ -1,4 +1,4 @@
-//go:build !release
+//go:build subprocess_computeprovider && !release
```
02strich:
```suggestion
//go:build computeprovider_subprocess && !release
```

02strich (approved):

wci/workflow/compute_provider/subprocess.go
johanngan: Changed

PR #130, opened 2026-09-11
Check workercontroller.enabled in more places

johanngan:
## What was changed
Change the version workflow and validate spec workflow to bail early on various operations if the `workercontroller.enabled` feature flag is off.

## Why?
Right now `workercontroller.enabled` isn't fully respected; the worker controller version/validate workflows will run mostly unimpeded even if the flag is false.


johanngan:
I don't really like this approach, but I couldn't think of anything better. Happy to consider alternative suggestions.

wci/workflow/workflow.go
```diff
@@ -120,7 +133,23 @@ func Workflow(ctx workflow.Context, unsafeWorkflowVersionGetter func() WorkerCon
...
+		// Since the setting is from dynamic config, it could change at any time, so the
+		// workflow itself needs to poll it periodically to be able to react.
+		//
+		// The timeout is set to 5 minutes, the same as the PullStats maxPollInterval.
+		// This will create a new workflow run (via CaN) per WCI every 5 minutes, even if WCI
+		// is disabled, but this is no worse than what would happen if it were enabled because
+		// the workflow would CaN at least every 5 minutes anyway after each PullStats.
+		// It's acceptable that the workflow may take up to 5 minutes to notice a config change.
+		workflow.Sleep(ctx, workerControllerEnabledPollInterval)
+		return workflow.NewContinueAsNewError(ctx, iface.WorkerControllerInstanceWorkflowType, workflowRunner.WorkerControllerInstanceWorkflowArgs)
```
justinschoeff: Wouldn't this create an infinite loop of: not enabled -> sleep -> CAN -> not enabled..

justinschoeff (commented):
What is driving this change? I feel like the [code that starts the workflow](https://github.com/temporalio/temporal-auto-scaled-workers/blob/main/wci/client/client.go#L311-L362) should be gated by this, not the workflow code itself? I think that would also allow us to [return an error](https://github.com/temporalio/temporal/blob/main/service/worker/workerdeployment/client.go#L2146-L2147) to the API/user.

johanngan:
> What is driving this change? I feel like the [code that starts the workflow](https://github.com/temporalio/temporal-auto-scaled-workers/blob/main/wci/client/client.go#L311-L362) should be gated by this, not the workflow code itself? I think that would also allow us to [return an error](https://github.com/temporalio/temporal/blob/main/service/worker/workerdeployment/client.go#L2146-L2147) to the API/user.

I'll follow up with you offline about what's driving this change.

If we gate the code that starts the workflow, then workflows that are already running for existing WDVs will continue to run, irrespective of the flag. I don't think that's desirable?

wci/workflow/workflow.go
johanngan: That was the idea, yeah. The point is that if we want the flag to have an effect on existing, already-running workflows, we need to do some sort of polling on it in a workflow-safe way.

The workflow already runs in an infinite loop with CaN anyway, so this wouldn't change that aspect of the workflow.

justinschoeff:
> > What is driving this change? I feel like the [code that starts the workflow](https://github.com/temporalio/temporal-auto-scaled-workers/blob/main/wci/client/client.go#L311-L362) should be gated by this, not the workflow code itself? I think that would also allow us to [return an error](https://github.com/temporalio/temporal/blob/main/service/worker/workerdeployment/client.go#L2146-L2147) to the API/user.
> 
> I'll follow up with you offline about what's driving this change.
> 
> If we gate the code that starts the workflow, then workflows that are already running for existing WDVs will continue to run, irrespective of the flag. I don't think that's desirable?

Makes sense. Either here or as a follow-up, we should update the start workflow to error up front as a UX improvement. 

I wonder if there is a way to surface an error to the UI/API/WDV state. For existing users that haven't enabled this config, but have been successfully using WCI/serverless, their workloads will just stop working without a way to debug in their existing workflows. If serverless was working, I updated temporal and then it stopped, i would think its a bug. I might go and recreate the worker deployment version or a new one and see it succeed, but not work it would reinforce the idea that there is a bug in update.

justinschoeff (approved):

wci/workflow/workflow.go
02strich: I don't think it matters with the existing workflows. The question is to prevent new ones as that was the intend. We should stop creation in the client or the API layer IMHO.

What is the scenario where you see doing this for existing workflows matter?

johanngan: Per offline discussion with Stefan, changed to a hybrid approach:
- Check the flag client-side for user-friendly error messages during WDV creation/update
- Check the flag server-side in update handlers (and others, TBD) in the version workflow, and at the start of the validate workflow

johanngan:
Reviewers: The three commits in this PR are clean to view and mostly independent. Version workflow checks, validate workflow checks, and client-side checks.

Also see: https://github.com/temporalio/temporal/pull/12180, which is required for the error message about WCI being disabled to actually make it to end users trying to configure WDVs under default server config.

Tested locally by building it into Temporal server alongside https://github.com/temporalio/temporal/pull/12180, trying to create a new/update an existing WDV with compute config via the CLI and the Go SDK, and observing the outputs for both values of `workercontroller.enabled`.

PR #132, opened 2026-09-14
Enforce max timeout for AgentCore invoke call

justinschoeff:
The AgentCore API is synchronous and while it does support streaming
responses, that is left up to the runtime code itself. This means the
invocation request can hang up to 8 hours!

Even when runtimes last a minute it is problematic for WCI. This commit
aims to limit the allowed runtime to 5 seconds and assuming success.

## What was changed
Add a timeout via context to the AWS AgentCore SDK InvokeAgentRuntime call to prevent it from hitting the activity timeout. This also adds an http trace on request write as a gauge to whether the invocation was actually transmitted or not. 

## Why?
Without this, the runtime can hit the activity timeout causing retries and create worker-deployment failures. Additionally, each call (including retries) does spin up a new worker agentcore side.

## Checklist

1. Closes 

2. How was this tested:
Launched an agent runtime that exceeded the 60s activity timeout and saw success alongside a consistent 5s between activity start to completed (hitting the agentCoreMaxRuntime.

Launched an agent runtime that responded immediately and saw success.

3. Any docs updates needed?


wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -17,6 +20,13 @@ import (
 	"go.temporal.io/server/common/dynamicconfig"
 )
 
+// AgentCore has no async invocation mode and it's 8hr session window means it doens't have to send headers until then.
```
02strich: declaring all timeouts as success has the risk of hiding upstream issues - what is our scheme there? I believe we should at least consider marking the connection unhealthy

wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -137,15 +147,37 @@ func (p *awsAgentCoreComputeProvider) invokeWorker(ctx context.Context, rc Reque
 		Qualifier:       aws.String(params.EndpointName),
 	}
 
-	// AgentCore doesn't have an async invocation method. This aims to make the call and close the response stream on
-	// our end as soon as possible, preventing this activity from blocking.
-	resp, err := invokeClient.InvokeAgentRuntime(ctx, input)
+	// Setup agentCore ctx with timeout to prevent extended blocking
+	invokeCtx, cancel := context.WithTimeout(ctx, agentCoreMaxRuntime)
+	defer cancel()
```
02strich, line 152: so what are you planning to capture with this vs the SDK timeouts?

wci/workflow/compute_provider/aws_agentcore.go
johanngan: I wonder if there's even a way to do this properly on the WCI side without getting an initial response from the `InvokeAgentRuntime` request. It seems like the agentcore SDK might not expose a way to check on the status of a specific session or session invocation, and if that's true then there'd be no other way for us to know whether the invoke succeeded (beyond side effects the worker has like polling task queues).

This makes me wonder if we'd have to enforce streaming responses on the AgentCore Runtime side through dedicated SDK helpers to run AgentCore workers, similar to how we have `lambdaworker` for Lambda.

justinschoeff: For comparison, lambda async invoke gives us confidence the lambda will be invoked, but not that it started or succeed. That would be ideal, and is what this aims to get close to.

This PR is doing two things:
- a context level timeout, tighter than the activity's. This allows the SDK to retry on it's timeouts like a connect or tls handshake and still succeed overall. Additionally, this context produces a ctx deadline error which we can distinguish from the SDK timeouts. So if the SDK times out we can bubble it up as a timeout/error and activity can retry or workflow can surface it.
- A http client trace tracking request write. After connect and request is sent, we set a boolean denoting request sent. This is a bit funky, but it gives us some more indication that we hit our deadline after the sdk sent the request.

justinschoeff: *for clarity, unless im missing something, this does not hide any SDK timeouts and will not consider them success. The constraint for success:
-  the Runtime returned before maxAgentRuntime successfully
- DeadlineExceed (NOT SDK Timeout) AND the call hit the WriteRequest step

02strich: I think it might be good to mention the difference in this comment so it doesn't require reading as much code

justinschoeff: Updated comment

wci/workflow/compute_provider/aws_agentcore.go
justinschoeff: I think this is answered in other comment.

wci/workflow/compute_provider/aws_agentcore.go
```diff
@@ -17,6 +20,13 @@ import (
 	"go.temporal.io/server/common/dynamicconfig"
 )
 
+// AgentCore has no async invocation mode and its blocking mechanism can max out our activity timeout, causing activity
+// retries and WDV failures. While customer config on the Runtime side can mitigate this, we can't rely on that.
+// Instead we combine this agentCoreMaxRuntime timeout with an HTTP trace giving us confidence that the invocation was
+// triggered. This is similar to Lambda's async invoke, we aren't concerned with the runtime operating successfully
+// beyond invocation. This timeout is the max time we'll wait for any runtime invocation.
+var agentCoreMaxRuntime = 5 * time.Second
```
mahars80, line 28: Does this feel any aggressive? Also, a noob question, why does this not meet the bar to be a config that can be changed at runtime? :)

justinschoeff: It is a bit aggressive. The only way this really becomes a problem is if the invoke call from Temporal server  to AWS takes 5s before sending the request, which shouldn't be the case and usually would be transient. The activity here would retry and hopefully succeed. I'm not completely opposed to making configurable server-side (not WDV/user configurable), but I'm not sure it's worth it.

02strich (approved):

mahars80 (approved):

